PR #6, opened 2021-08-26
[Bernard Boey] iP

bernardboey:


src/main/java/Duke.java
```diff
@@ -1,10 +1,136 @@
...
     public static void main(String[] args) {
-        String logo = " ____        _        \n"
-                + "|  _ \\ _   _| | _____ \n"
-                + "| | | | | | | |/ / _ \\\n"
-                + "| |_| | |_| |   <  __/\n"
-                + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
+        printGreeting();
+        handleCommands();
+        printFarewell();
```
markuslyq: Great job on refactoring everything into smaller functions.

src/main/java/Duke.java
```diff
@@ -1,10 +1,136 @@
...
+    /** Number of tasks */
+    private static int taskCount = 0;
+
+    private static String indent(String text) {
+        String[] lines = text.split(System.lineSeparator());
+        for (int i = 0; i < lines.length; i++) {
+            lines[i] = " ".repeat(5) + lines[i];
+        }
+        return String.join(System.lineSeparator(), lines);
+    }
```
Isabella-L: Perhaps this can be changed to `System.out.println("/t")`?

src/main/java/Event.java
```diff
@@ -0,0 +1,29 @@
...
+        this.at = at;
+    }
+
+    /**
+     * Returns a string representation of the event (consisting of the symbol for the task type, the event date and
+     * time, as well as the string representation from the parent class).
+     *
+     * @return A string representation of the event.
+     */
+    @Override
```
Isabella-L: I like how you do detailed documentation

src/main/java/Deadline.java
```diff
@@ -0,0 +1,29 @@
...
+public class Deadline extends Task {
+    /** Deadline for the task (date/time) */
+    protected String by;
+
+    /**
+     * Creates a task with the specified description and deadline.
+     *
+     * @param description Description of the task.
+     * @param by Deadline for the task.
+     */
```
serenakemono: I like the descriptions of the methods.

src/main/java/Duke.java
```diff
@@ -1,10 +1,136 @@
...
+
+    private static void printGreeting() {
+        String greeting = "Hello! I'm Duke" + System.lineSeparator()
+                + "What can I do for you?";
+        printResponseBlock(greeting);
+    }
+
+    /** Prints out the list of tasks (numbered) together with their status icons */
+    private static void printTasks() {
+        String[] formattedTasks = new String[taskCount];
```
serenakemono: I like the naming of this variable which clearly distinguishes itself from normal tasks.

src/main/java/Duke.java
```diff
@@ -1,10 +1,136 @@
...
+        for (int i = 0; i < taskCount; i++) {
+            formattedTasks[i] = String.format("%d.%s", i + 1, tasks[i]);
+        }
+        String formattedTaskList = String.join(System.lineSeparator(), formattedTasks);
```
serenakemono: I'm not sure about this variable name as it is actually not a "list" but a string. Perhaps "formattedTaskString" would be less confusing?

src/main/java/Duke.java
```diff
@@ -1,10 +1,136 @@
...
+
+    private static void addTask(Task task) {
+        tasks[taskCount] = task;
+        taskCount++;
+        printResponseBlock("Got it. I've added this task:" + System.lineSeparator()
+                + "  " + task + System.lineSeparator()
+                + "Now you have " + taskCount
+                + (taskCount > 1
+                ? " tasks in the list"
+                : " task in the list")
```
serenakemono: I like the clear formatting here.

serenakemono (commented):
Overall I really like the strict adherence to the coding standard in your code. Good job! :)

src/main/java/Duke.java
```diff
@@ -1,10 +1,136 @@
...
+     * Horizontal lines will be printed before and after the specified text, and the text will be indented.
+     *
+     * @param text Text to be printed out.
+     */
+    private static void printResponseBlock(String text) {
+        System.out.println(INDENTED_HORIZONTAL_LINE);
+        System.out.println(indent(text));
+        System.out.println(INDENTED_HORIZONTAL_LINE);
+        System.out.println();
+    }
```
markuslyq: I like that there is a function for the output messages so as to standardise all the output.

src/main/java/Duke.java
```diff
@@ -1,10 +1,136 @@
...
+        printResponseBlock("Nice! I've marked this task as done:" + System.lineSeparator()
+                + formattedTask);
+    }
+
+    private static void addTask(Task task) {
+        tasks[taskCount] = task;
+        taskCount++;
+        printResponseBlock("Got it. I've added this task:" + System.lineSeparator()
+                + "  " + task + System.lineSeparator()
+                + "Now you have " + taskCount
```
markuslyq: I like that you start on a new line after every line separator, it makes the code much more readable.

src/main/java/Task.java
```diff
@@ -0,0 +1,53 @@
...
+    /**
+     * Returns the status icon for the task (based on whether it is completed).
+     * The icon is a string of length 1 ("X" if completed, or " " if not).
+     *
+     * @return Status icon for the task.
+     */
+    public String getStatusIcon() {
+        return (isDone ? "X" : " ");
+    }
+
```
markuslyq: Great description of methods

markuslyq (commented):
Overall, really well done and the code strictly adheres to the coding standards.

src/main/java/Event.java
bernardboey: Thank you!

src/main/java/Duke.java
bernardboey: Thanks!

src/main/java/Task.java
bernardboey: Thanks!

src/main/java/Duke.java
bernardboey: Yup indeed it is to standardise the output! Thank you :)

src/main/java/Duke.java
bernardboey: That's what I was going for, thank you!

src/main/java/Deadline.java
bernardboey: Thanks!

src/main/java/Duke.java
bernardboey: Thanks!

src/main/java/Duke.java
bernardboey: That's true. I prefer not to use `formattedTaskString` as it may imply that it refers to only one task. Perhaps `taskListOutput` instead?

src/main/java/Duke.java
bernardboey: That's true. I prefer using spaces so that I can control the amount of indentation (e.g. horizontal line indented by 4 spaces, text indented by 5 spaces). Thanks for your suggestion though!

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,190 @@
...
+    /**
+     * Executes the command specified by the input.
+     *
+     * @param userInput Input command together with any arguments.
+     * @return Feedback about what was executed.
+     */
+    private static String executeCommand(String userInput) {
+        final String[] commandAndArgs = userInput.split(" ", 2);
+        final String command = commandAndArgs[0];
+        final String args = commandAndArgs.length > 1 ? commandAndArgs[1] : "";
```
kaiwen98: Replace magic numbers with well named constants.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,190 @@
...
+import duke.task.Task;
+import duke.task.Todo;
+
+/**
+ * A personal assistant chatbot.
+ */
+public class Duke {
+    private static final String INDENTED_HORIZONTAL_LINE = " ".repeat(4) + "_".repeat(60);
+    /** Platform independent line separator */
+    private static final String LS = System.lineSeparator();
```
kaiwen98: Suggest to avoid short-form variable names; replace the name with a more verbose one.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,190 @@
...
+
+    private static final Scanner SCANNER = new Scanner(System.in);
```
kaiwen98: Suggest to not use all-caps style on this variable name, for it is not a constant.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,190 @@
...
+    private static final Task[] tasks = new Task[CAPACITY];
+    /** Total number of tasks */
+    private static int taskCount = 0;
+
+    /**
+     * Main entry point of Duke.
+     */
+    public static void main(String[] args) {
+        printGreeting();
+        while (true) {
```
kaiwen98: Suggest to relocate your exit condition to this while loop for better logic clarity.

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    /**
+     * Returns the status icon for the task (based on whether it is completed).
+     * The icon is a string of length 1 ("X" if completed, or " " if not).
+     *
+     * @return Status icon for the task.
+     */
+    public String getStatusIcon() {
+        return (isDone ? "X" : " ");
```
kaiwen98, line 29: Suggest to keep consistent the reference of attributes, between this.isDone or isDone.

bernardboey: That's true, I've changed all to `isDone`. Thanks!

src/main/java/duke/Duke.java
bernardboey: Could I clarify, what do you mean by "not a constant"? I thought it was a constant because they are not changed once it is defined. I was also referring to the Contacts repository (https://github.com/nus-cs2113-AY2122S1/contacts/blob/master/src/main/java/Contacts1.java#L24 and https://github.com/nus-cs2113-AY2122S1/contacts/blob/master/src/main/java/Contacts1.java#L107)

src/main/java/duke/Duke.java
bernardboey: Done, changed to `LINE_SEPARATOR`.

src/main/java/duke/Duke.java
bernardboey: Done, thanks!

src/main/java/duke/Duke.java
bernardboey: Will implement this in Week 7 as relocating the exit condition requires some refactoring and it makes more sense to do it in conjunction with the stretch goal for `A-MoreOOP`. Thanks!

PR #10, opened 2021-08-27
[Swati] iP

swatimahadevan:


src/main/java/Duke.java
```diff
@@ -1,10 +1,118 @@
...
+        if (input.equals("list")) {
+            printTasksList();
+        }
+        else if (wordsInInputString[0].equals("done")) {
+            int taskIndex = Integer.parseInt(wordsInInputString[1]);
+            markTaskAsDone(taskIndex);
+        }
+        else {
+            addTaskToList(input);
+        }
```
AnShengLee:
```suggestion
        } else if (wordsInInputString[0].equals("done")) {
            int taskIndex = Integer.parseInt(wordsInInputString[1]);
            markTaskAsDone(taskIndex);
        } else {
            addTaskToList(input);
        }
```

src/main/java/Duke.java
```diff
@@ -1,10 +1,118 @@
+import java.util.ArrayList;
+import java.util.Scanner;
+
 public class Duke {
+    private static ArrayList<Task> tasksList = new ArrayList<>();
```
AnShengLee: Plural form should be used on names representing a collection of objects. For example: tasks

src/main/java/Task.java
```diff
@@ -0,0 +1,27 @@
+public class Task {
+
+    protected String description;
+    protected boolean isDone;
```
AnShengLee: Good boolean variable name!

src/main/java/Duke.java
```diff
@@ -1,10 +1,118 @@
...
+        printDividerLine();
+        System.out.println("Here are the tasks in your list:");
+        int i = 1;
+        for (Task item: tasksList) {
+            System.out.println((i++) + ". " + item);
+        }
+        printDividerLine();
+    }
+
+    public static void markTaskAsDone(int taskIndex) {
```
AnShengLee: Method name is a verb and is in camelCase. Good work.

AnShengLee (commented):
Overall, coding standard is followed with only a small mistake with a if-else block in Duke.java. Also, plural form should be used on names representing a collection of objects.

src/main/java/Duke.java
lixiyuan416: Good naming as you are using an arraylist. I respectfully disagree with An Sheng.

src/main/java/Duke.java
```diff
@@ -1,10 +1,118 @@
...
+        printDividerLine();
+    }
+
+    public static void printByeMessage() {
+        printDividerLine();
+        System.out.println("Bye!");
+        printDividerLine();
+    }
+
+    public static void addTaskToList(String task) {
```
lixiyuan416: Beautifully coded whereby method calls are at the same level of abstraction. Good use of extracting and refactoring

src/main/java/Duke.java
```diff
@@ -1,10 +1,118 @@
...
+        System.out.println("Now you have " + tasksList.size() + " tasks in the list.");
+        printDividerLine();
+    }
+
+    private static void addToDo(String input) {
+        input = input.split("todo")[1].trim();
+        tasksList.add(new ToDo(input));
+    }
+
+    private static void addEvent(String input) {
```
lixiyuan416: User input processing code is easy to understand, good job!

src/main/java/Duke.java
```diff
@@ -1,10 +1,118 @@
...
+        String by = input.split("/by")[1].trim();
+        tasksList.add(new Deadline(description, by));
+    }
+
+    public static void printTasksList() {
+        printDividerLine();
+        System.out.println("Here are the tasks in your list:");
+        int i = 1;
+        for (Task item: tasksList) {
+            System.out.println((i++) + ". " + item);
```
lixiyuan416: Consider incrementing i on another line so readers don't have to wonder "is your first item 1 or 2?"

src/main/java/Duke.java
```diff
@@ -1,10 +1,118 @@
...
     public static void main(String[] args) {
-        String logo = " ____        _        \n"
-                + "|  _ \\ _   _| | _____ \n"
-                + "| | | | | | | |/ / _ \\\n"
-                + "| |_| | |_| |   <  __/\n"
-                + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
-    }
-}
+        Scanner in = new Scanner(System.in);
```
lixiyuan416: Consider naming your Scanner variable userInput

lixiyuan416 (commented):
Very neat coding style, keep it up!

src/main/java/Duke.java
```diff
@@ -1,10 +1,208 @@
...
+        try {
+            if (task.startsWith("deadline")) {
+                addDeadline(task);
+            } else if (task.startsWith("event")) {
+                addEvent(task);
+            } else if (task.startsWith("todo")) {
+                addToDo(task);
+            } else {
+                handleUnknownCommand(task);
+            }
```
sinteary: Good SLAP here

src/main/java/Duke.java
```diff
@@ -1,10 +1,208 @@
...
-                + "| | | | | | | |/ / _ \\\n"
-                + "| |_| | |_| |   <  __/\n"
-                + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
-    }
-}
+        Scanner in = new Scanner(System.in);
+        printGreetMessage();
+        String input;
+        try{
```
sinteary: There should be a space here between `try` and `{`

src/main/java/Storage.java
```diff
@@ -0,0 +1,84 @@
...
+        fw.close();
+    }
+
+    public static ArrayList<Task> loadData() throws FileNotFoundException{
+        File directory = new File(SAVEDIR);
+        if(!directory.exists()){
+            if(directory.mkdir()){
+                System.out.println("Successfully created a new directory for save file");
+            }
+        } else {
```
sinteary: You can try to make the [happy path](https://nus-cs2113-ay2122s1.github.io/website/se-book-adapted/chapters/codeQuality.html#make-the-happy-path-prominent) more prominent to avoid nesting these conditionals. Eg. if directory exists and direction.mkdir(), `return`. Then the rest of your code does not need to be nested in the `else` block

src/main/java/Storage.java
```diff
@@ -0,0 +1,84 @@
...
+import java.util.Scanner;
+
+public class Storage {
+    private static ArrayList<Task> tasks = new ArrayList<>();
+    private static final String FILEPATH = "data/savedData.txt";
+    private static final String SAVEDIR = "data";
+
+    public static void saveData(ArrayList<Task> tasks) throws IOException{
+        StringBuilder line = new StringBuilder();
+        for(Task task: tasks){
```
sinteary: Need to add spaces between `for` and the bracketed condition, and the `{`, likewise for other similar parts of the code with control flow statements

src/main/java/Storage.java
```diff
@@ -0,0 +1,84 @@
...
+        String[] splitTask = serializedTask.split("\\|");
+
+        String taskType = splitTask[0];
+        boolean isDone = splitTask[1].equals("1");
+        String description = splitTask[2].trim();
+
+        switch (taskType){
+            case "T":
+                task = new ToDo(description, isDone);
+                break;
```
sinteary: This does not meet the coding standard as there should be no indentation for `case` statements

src/main/java/Storage.java
```diff
@@ -0,0 +1,84 @@
...
+                task = new Deadline(description, by, isDone);
+                break;
+            case "E":
+                String at = splitTask[3].trim();
+                task = new Event(description, at, isDone);
+                break;
+        }
+        return task;
+    }
+    private static String serializeTask(Task task){
```
sinteary: You should leave a line here between function definitions for better readability

src/main/java/Duke.java
```diff
@@ -1,10 +1,208 @@
...
+            int index;
+            try {
+                index = Integer.parseInt(input.split("done")[1].trim());
+                markTaskAsDone(index);
+            } catch (DukeInvalidArgumentException e) {
+                printDividerLine();
+                System.out.println("☹ OOPS!!! No such item in the list :-(");
+                printDividerLine();
+            }
+        } else if (input.startsWith("delete")){
```
sinteary: Including the logic inside this function would cause the function to be quite long (especially if the list of commands grows). Would recommend extracting out into smaller functions that handle each type of command

PR #14, opened 2021-08-27
[Christopher Nge] iP

Poopies99:
:)


src/main/java/Duke.java
```diff
@@ -1,10 +1,134 @@
...
+                + "    ____________________________________________________________\n";
+        toDo[trackIndex] = new Task(line);
+        System.out.println(output);
+    }
+
+    private static void handleEventRequest(String line, Task[] toDo, int trackIndex) {
+        String output;
+        int dividerPosition = line.indexOf("/");
+        String description = line.substring(6, dividerPosition - 1);
+        String date = line.substring(dividerPosition + 4);
```
crabnuggets: Perhaps you can consider using `indexOfFoo` etc. instead of 6 (line 51) and 4 (line 52) to avoid the use of magic literals. This occurs in other handle request functions as well such as `handleDeadlineRequest`.

src/main/java/Duke.java
```diff
@@ -1,10 +1,134 @@
...
+                + "      " + toDo[trackIndex].toString() + "\n"
+                + "     Now you have " + (trackIndex + 1)  + " tasks in the list.\n"
+                + "    ____________________________________________________________";
+        System.out.println(output);
+    }
+
+    private static void handleDoneRequest(String line, Task[] toDo) {
+        String output;
+        int dividerPosition = line.indexOf(" ");
+        int numberInTaskArray = Integer.parseInt(line.substring(dividerPosition + 1)) - 1;
```
crabnuggets: Perhaps you can consider breaking this line down into smaller bits as the expression here is rather long.

src/main/java/Task.java
```diff
@@ -0,0 +1,25 @@
...
+        this.description = description;
+        this.isDone = false;
+    }
+
+    public String toString() {
+        return getStatusIcon() + getDescription();
+    }
+
+    public String getStatusIcon() {
+        return "[" + (isDone ? "X" : " ") + "] ";
```
crabnuggets: Perhaps the `isDone` ternary operation can be extracted into a new line to simplify the return expression.

src/main/java/Deadline.java
```diff
@@ -0,0 +1,26 @@
+public class Deadline extends Todo {
+    protected String by;
```
crabnuggets: `by` could come off as an ambiguous/awkward sounding field name. Perhaps something clearer like "date" or "time" would be better.

src/main/java/Duke.java
```diff
@@ -1,10 +1,134 @@
...
+            } else if (line.contains("deadline")) {
+                handleDeadlineRequest(line, toDo, trackIndex);
+                trackIndex++;
+            } else if (line.contains("event")) {
+                handleEventRequest(line, toDo, trackIndex);
+                trackIndex++;
+            } else {
+                handleTaskRequest(line, toDo, trackIndex);
+                trackIndex++;
+            }
```
crabnuggets: I like how neat this entire if-else block is and how it follows coding standards and is an example of good code quality.

crabnuggets (commented):
Looks good! 
Some comments about coding standards have been given.

src/main/java/Duke.java
```diff
@@ -34,19 +31,51 @@ public static void main(String[] args) {
...
+            } else if (line.contains("deadline")) {
+                int dividerPosition = line.indexOf("/");
+                String description = line.substring(9, dividerPosition - 1);
+                String date = line.substring(dividerPosition + 4);
+                toDo[trackIndex] = new Deadline(description, date);
+                output = "    ____________________________________________________________\n"
+                        + "     Got it. I've added this task:\n"
+                        + "      " + toDo[trackIndex].toString() + "\n"
+                        + "     Now you have " + (trackIndex + 1) + " tasks in the list.\n"
+                        + "    ____________________________________________________________";
```
KaiserHuang88: Maybe can write a separate function for writing output? So as to not make each output assignment so long.

src/main/java/Duke.java
```diff
@@ -34,19 +31,51 @@ public static void main(String[] args) {
...
+                trackIndex++;
+            } else if (line.contains("event")) {
+                int dividerPosition = line.indexOf("/");
+                String description = line.substring(6, dividerPosition - 1);
```
KaiserHuang88: Probably shouldn't use magic numbers like 6 and 4

src/main/java/Deadline.java
```diff
@@ -0,0 +1,26 @@
...
+        return by;
+    }
+
+    public void setBy(String outcome) {
+        by = outcome;
+    }
+
+    @Override
+    public String toString() {
+        return super.toString() + " (by: " + getBy() + ")";
```
KaiserHuang88: This line is a little vague, can explain more.

src/main/java/Duke.java
```diff
@@ -15,15 +15,12 @@ public static void main(String[] args) {
...
         Scanner in = new Scanner(System.in);
 
-        //Initialize List
-        Task[] toDo = new Task[10];
+        Task[] toDo = new Task[100];
         int trackIndex = 0;
-        //Initialize bool to check whether program should exit loop
         boolean isProgress = true;
 
         while (isProgress) {
```
KaiserHuang88, line 30: Maybe can just use while (true)? Then put a break line somewhere.

KaiserHuang88 (commented):
Very well written code! Only a few nitpicks here and there.

src/main/java/Duke.java
Poopies99: Yeap, you're right! Thanks for the tip!

src/main/java/Duke.java
Poopies99: Yeap you're right! Thanks for the tip!

PR #15, opened 2021-08-27
[Alvin Tan] iP

alvintan01:


src/main/java/Duke.java
```diff
@@ -1,10 +1,230 @@
...
+        if (currentTask.isDone()) {
+            customPrint("You have already completed the task [" + currentTask.getDescription() + "]!");
+        } else {
+            currentTask.markAsDone(); // Mark current task as done
+            String textToPrint = "Nice! I've marked this task as done:\n";
+            textToPrint += currentTask.toString();
+            customPrint(textToPrint);
+        }
     }
 }
```
itsleeqian, line 59: Code has been modularized well. Variable and Function names have been named in camelCase. 
Numerous comments are also good. Good job!

src/main/java/Duke.java
```diff
@@ -1,10 +1,230 @@
...
+
+    /**
+     * Marks a task as done and calls customPrint to print the completed message.
+     *
+     * @param description String ID of task to be marked as completed.
+     */
+    public static void markDone(String description) {
+        int taskId = Integer.parseInt(description) - 1; // -1 as array index starts from 0
+
+        // Checks for valid taskID
```
itsleeqian: Perhaps you can create another boolean variable to store the whole line, i.e isValidTaskID

alvintan01: Thanks for your feedback!

src/main/java/Duke.java
```diff
@@ -2,97 +2,187 @@
...
+            // Splits user input by spaces
+            String[] userInputSplited = userInput.split("\\s+", 2);
+            String command = userInputSplited[0];
+            String description = "";
+
+            // Ensure description exists
+            if (userInputSplited.length > 1) {
+                description = userInputSplited[1];
             }
+            processCommands(command, description);
```
nguyin: Good job changing if-else statements in main function to switch-case in processCommands function, I think it's better to avoid the deeply nested code

src/main/java/Duke.java
```diff
@@ -2,97 +2,187 @@
...
+     *
+     * @param description Task description entered by user.
+     */
+    private static void addEvent(String description) {
+        int byIndex = description.indexOf("/at");
+        String taskDescription = description.substring(0, byIndex).trim(); // Remove trailing spaces
+        String at = description.substring(byIndex + 4).trim(); // Remove leading spaces
+        tasks.add(new Event(taskDescription, at));
+        acknowledgeCommand();
+    }
```
nguyin: Good job with keeping all methods short but effective

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,363 @@
...
+        customPrint(taskString);
+    }
+
+    /**
+     * Marks a task as done and calls customPrint to print the completed message.
+     *
+     * @param id String ID of task to be marked as completed.
+     */
+    public static void markDone(String id) {
+        if (!checkValidTaskId(id)) { // Task id is invalid
```
OngDeZhi: You may want to consider changing the method name called, checkValidTaskId to ValidTaskId instead, it would read better.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,363 @@
...
+    private static void exit() {
+        customPrint("Bye. Hope to see you again soon!");
+        // Ends the program
+        System.exit(0);
+    }
+
+    /**
+     * Prints the acknowledgement that the command is processed.
+     */
+    private static void acknowledgeCommand() {
```
OngDeZhi: This seems to not acknowledge the command that is processed, but it acknowledges the add command that is processed.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,363 @@
...
+        String[] output = {taskDescription, date};
+        return output;
+    }
+
+    /**
+     * Adds a Deadline to the Task list
+     *
+     * @param description Task description entered by user.
+     */
+    private static void addDeadLine(String description) {
```
OngDeZhi: Should be in camelCase, don't have to capitalize the L.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,363 @@
...
+        String welcomeMessage = logo;
+        welcomeMessage += "Hello! I'm Duke\n";
+        welcomeMessage += "What can I do for you?";
+        customPrint(welcomeMessage);
+    }
+
+    /**
+     * Prints the invalid command message.
+     */
+    private static void printInvalidCommand() {
```
OngDeZhi: May want to rename this method, since it is not printing the invalid command, but the invalid command message.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,363 @@
...
+            throw new InvalidCommand();
+        }
+    }
+
+    /**
+     * Adds a Todo to the Task list
+     *
+     * @param description Task description entered by user.
+     */
+    private static void addToDo(String description) {
```
OngDeZhi: Should be in camelCase, don't have to capitalize the D.

PR #18, opened 2021-08-27
[Alvin Ben Abraham] iP

alvynben:
<div align="center">

# Do enjoy!

</div>


src/main/java/Event.java
```diff
@@ -2,16 +2,29 @@
...
 
+    /**
+     * Creates a deadline with task type [E], and timing based on user's input.
+     * @param name The name of the event.
+     * @param timing The timing of the event.
+     */
     public Event(String name, String timing) {
         super(name);
         this.timing = timing;
         this.typeOfTask = "[E]";
```
Cuiminjing: I think for typeOfTask, dont need  this.typeOfTask

src/main/java/Event.java
```diff
@@ -2,16 +2,29 @@
...
         return timing;
     }
 
+    /**
+     * Returns the event in a human-readable format.
+     * @return [Type of Task][Completion Status] [Name of Task] ([Timing of Task])
+     */
     @Override
     public String printTask() {
         return this.getTypeOfTask() + this.getDoneStatusAsSymbol() + " "
```
Cuiminjing: maybe you can conclude these two lines in one

src/main/java/Task.java
```diff
@@ -56,7 +56,7 @@ public void setDone(boolean done) {
     }
```
Cuiminjing: for line 38: i think we donot need a function to get the type of this task because once it get into this function, it should be an event. hence, we only need tyo return [E] when printing

src/main/java/Todo.java
```diff
@@ -11,7 +11,7 @@ public Todo(String name) {
     }
 
     /**
-     * Prints task in a human-readable format.
+     * Returns task in a human-readable format.
      * @return [Type of Task][Completion Status] [Name of Task]
      */
     public String printTask() {
```
Cuiminjing: Overall， I think you did a great job in code standards!!!: :)

src/main/java/Event.java
alvynben: Thanks, will take that into consideration!

src/main/java/Event.java
alvynben: Thanks for the suggestion! I think I'd still prefer it this way in case I want to get the TypeOfTask and DoneStatusAsSymbol separately in the future. Please do let me know if you disagree!

src/main/java/Task.java
alvynben: Wouldn't getTypeOfTask still be useful when I need to print out a list of tasks? That was my main intention when I made this. Do let me know if I'm missing something!

src/main/java/Todo.java
alvynben: Thank you very much :)

src/main/java/triss/Triss.java
```diff
@@ -0,0 +1,429 @@
...
+            "   ██    ██   ██ ██ ██      ██      \n" +
+            "   ██    ██████  ██ ███████ ███████ \n" +
+            "   ██    ██   ██ ██      ██      ██ \n" +
+            "   ██    ██   ██ ██ ███████ ███████ \n";
+
+    /** String of underscores to separate user input and Triss output */
+    public static final String SEPARATOR_LINE = "____________________________________________________________";
+
+    /** Boolean to track if user has said "bye" */
+    private static boolean hasUserSaidBye = false;
```
Ashley-Lau: If these variables are specific to this Triss class/object, why not consider making all methods/some variables non static?

src/main/java/triss/Triss.java
```diff
@@ -0,0 +1,429 @@
...
+                printAllTasks();
+                break;
+            case "done":
+                handleUserMarkingTaskAsDone(userInput);
+                break;
+            case "delete":
+                handleUserDeletingTask(userInput);
+                break;
+            default:
+                try {
```
Ashley-Lau: Perhaps a better/ more clear implementation would be to have a case for the respective creatingTask commands so that the default case can throw an error straight.

src/main/java/triss/Triss.java
```diff
@@ -0,0 +1,429 @@
...
+     * Informs user if task was already done.
+     * @param userInput Any user input starting with "done"
+     */
+    private static void handleUserMarkingTaskAsDone(String userInput) {
+        // Get number of task after the term "done"
+        int indexOfCompletedTask;
+        // Throw exception if user did not type a number after "done"
+        try {
+            indexOfCompletedTask = Integer.parseInt(parseUserInput(userInput, 1)) - 1;
+        } catch (Exception e) {
```
Ashley-Lau: if you have a try catch block at a parent method that calls this method, you could throw an error here so and catch the error at the parent method so that you do not need empty returns or so many try catch blocks. (Occurred many times in your code) Essentially, 1 try catch is possible for the entire code.

src/main/java/triss/Triss.java
```diff
@@ -0,0 +1,429 @@
...
+        // Print out the task in the following format: "    [X] Task"
+        printLine("    " + chosenTask.printTask());
+    }
+
+    /**
+     * Prints all tasks stored in Task Array tasks.
+     */
+    private static void printAllTasks() {
+        // If user said "list", print a list of all saved tasks
+        for (Task task:tasks) {
```
Ashley-Lau: Minor nitpick but you could add spaces between the `:` for slightly better readability.

src/main/java/triss/exception/TrissException.java
```diff
@@ -0,0 +1,8 @@
+package triss.exception;
+
+public class TrissException extends Exception{
```
Ashley-Lau: Space between Exception and {

src/main/java/triss/Triss.java
```diff
@@ -0,0 +1,429 @@
...
+    }
+
+    /**
+     * Mark user task as done, if request is valid.
+     * Stops if user did not specify a task.
+     * Stops if user's chosen task does not exist.
+     * Informs user if task was already done.
+     * @param userInput Any user input starting with "done"
+     */
+    private static void handleUserMarkingTaskAsDone(String userInput) {
```
Ashley-Lau: The method name could be more concise. Maybe `markTaskAsDone` is sufficient?

Ashley-Lau (commented):
Overall, good effort for abstraction of methods and its funtionalities, documentation (although formatting can be improved) and checks for the input. You could work on the try-catch improvement comment i gave in one of the lines to improve code quality!

src/main/java/triss/Triss.java
alvynben: Thanks, will try to decrease number of static variables.

src/main/java/triss/Triss.java
alvynben: Roger that, thanks.

src/main/java/triss/Triss.java
alvynben: Do you think it's okay if I have try catch blocks that throw errors? The benefit is that I don't need to do multiple if-else statements to check the validity of the input, and then throw errors.

src/main/java/triss/Triss.java
alvynben: Roger that.

src/main/java/triss/exception/TrissException.java
alvynben: Nice catch! Thanks!

src/main/java/triss/Triss.java
alvynben: I wanted to standardise the naming here as they are user command handlers? does that make sense?

PR #27, opened 2021-08-27
[Joshua Dass] iP

JoshHDs:


src/main/java/Duke.java
```diff
@@ -1,10 +1,136 @@
+import java.util.Scanner;
+
 public class Duke {
+
+    public static void showWelcomeMessage() {
```
leowyy99: It would be better if this method is private, as it is not used in elsewhere. Same for all other methods in the class except for main().

src/main/java/Duke.java
```diff
@@ -1,10 +1,136 @@
...
+        System.out.println("Hello! I'm Duke");
+        System.out.println("What can I do for you?");
+    }
+
+    public static void printList(Task[] list, int size) {
+        int position = 1;
+        System.out.println("    Here are the tasks in your list:");
```
leowyy99: Consider using \t escape character

src/main/java/Duke.java
```diff
@@ -1,10 +1,136 @@
...
+        case "todo":
+            ToDos newToDo = new ToDos(command.substring(5));
+            addTask(list, newToDo, taskCount);
+            taskCount ++;
+            break;
+        case "deadline":
+            //extract out deadline description and by
+            Deadlines newDeadline = createNewDeadline(words);
+            addTask(list, newDeadline, size);
+            taskCount ++;
```
leowyy99: It would be better if there is no space characted between taskCount and the ++ operator.

src/main/java/Duke.java
```diff
@@ -1,10 +1,136 @@
...
+        String deadlineDescription = "";
+        for(int i = 1 ; i < words.length ; i ++) {
+            if(words[i].equals("/by")) {
+                deadlineIndex = i + 1;
+                by = words[deadlineIndex];
+                break;
+            }
+            deadlineDescription = deadlineDescription + " " + words[i];
+        }
+        Deadlines newDeadline = new Deadlines( deadlineDescription.substring(1), by);
```
leowyy99: May be better if there is no space character after the ( when calling a method, seems that is not consistent throughout the code.

src/main/java/Duke.java
```diff
@@ -1,10 +1,136 @@
...
+
+    public static void markTaskDone(Task[] list , String word) {
+        //convert string number to int number
+        int position = Integer.parseInt(word);
+        //if task does not exist, return
+        if (position > list.length) {
+            System.out.println("    Sorry! No such task!");
+            return;
+        }
+        //if task exist, mark task as done
```
leowyy99: I like the comments in the code, makes it easy to understand the flow. Good job on that!

src/main/java/Duke.java
```diff
@@ -41,6 +41,39 @@ public static void addTask( Task[] list, Task newTask, int size) {
...
+        }
+        Deadlines newDeadline = new Deadlines( deadlineDescription.substring(1), by);
+        return newDeadline;
+    }
+
+    public static Events createNewEvent(String[] words) {
+        int eventIndex = 0;
+        String timeAllocation = "";
+        String eventDescription = "";
+        for(int i = 1 ; i < words.length ; i ++) {
```
mohamad-adam8991: Perhaps this part of the code can be written to be less nested.

src/main/java/Duke.java
```diff
@@ -1,10 +1,136 @@
...
+    }
+
+    public static int distinguishCommand( String command , Task[] list, int size) {
+        //split into word array
+        String[] words = command.split(" ");
+        String firstWord = words[0];
+        int taskCount = size;
+
+        //determine command
+        switch (firstWord) {
```
mohamad-adam8991: I like the usage of functions to avoid deep nesting!

src/main/java/Duke.java
```diff
@@ -1,10 +1,136 @@
...
+
     public static void main(String[] args) {
-        String logo = " ____        _        \n"
-                + "|  _ \\ _   _| | _____ \n"
-                + "| | | | | | | |/ / _ \\\n"
-                + "| |_| | |_| |   <  __/\n"
-                + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
+
+        String line;
```
mohamad-adam8991: Perhaps a more informative variable name could be used here?

src/main/java/ToDos.java
```diff
@@ -0,0 +1,14 @@
...
+
+    //variables
+    public ToDos(String description) {
+        super(description);
+    }
+
+    //methods
+    @Override
+    public String toString() {
+        return( "[T]" + super.toString() );
```
mohamad-adam8991: Perhaps the spaces before and after the brackets could be omitted?

src/main/java/Duke.java
JoshHDs: Thank for pointing that out!

src/main/java/Duke.java
JoshHDs: Thanks for pointing that out!

src/main/java/Duke.java
JoshHDs: You are right! Will update that!

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,159 @@
...
+            break;
+        }
+
+        return taskCount;
+    }
+
+
+
+    public static void main(String[] args) {
+
```
kaiwen98, line 52: Suggest to remove the new line here to retain format consistency.

src/main/java/duke/Events.java
```diff
@@ -0,0 +1,21 @@
...
+        this.timeAllocation = timeAllocation;
+    }
+
+    //methods
+    @Override
+    public String toString() {
+        return("[E]" + super.toString() + " (at: " + timeAllocation + ")");
+    }
+
+
```
kaiwen98, line 31: Suggest to remove the unnecessary white spaces.

src/main/java/duke/Task.java
```diff
@@ -0,0 +1,40 @@
...
+        return description;
+    }
+    public boolean getDone() {
+        return isDone;
+    }
+
+    //setters
+    public void setDone() {
+        this.isDone = true;
+    }
```
kaiwen98, line 26: Suggest to change to getIsDone() and setIsDone() respectively, since you are modifying the variable isDone.

src/main/java/duke/Deadlines.java
```diff
@@ -0,0 +1,20 @@
+package duke;
+
+public class Deadlines  extends Task {
```
kaiwen98, line 12: Suggest to change class name to Deadline, since each instance of this class represents a single Deadline.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,159 @@
...
+    public static void printList(Task[] list, int size) {
+        int position = 1;
+        System.out.println("    Here are the tasks in your list:");
+        for( int i = 0 ; i < size ; i ++ ) {
+            System.out.println( "    " + position  + "." + list[i]);
+            position ++;
+        }
+    }
+
+    public static void markTaskDone(Task[] list , String word) {
```
kaiwen98: Suggest to change the variable name of "list" to "tasks" or "taskList", so as to remove ambiguity.

src/main/java/duke/ToDos.java
```diff
@@ -0,0 +1,17 @@
+package duke;
+
+import duke.Task;
+
+public class ToDos extends Task {
```
kaiwen98, line 8: Suggest to change class name to Todo, since each instance of this class represents a single Todo.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,159 @@
...
+                System.out.println("    OOPS!!! I'm sorry, but I don't know what that means :-(");
+
+            break;
+        }
+
+        return taskCount;
+    }
+
+
+
```
kaiwen98, line 50: Suggest to leave only one new line instead of 3, so as to maintain a consistent code format.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,159 @@
...
+        for(int i = 1 ; i < words.length ; i ++) {
+            if(words[i].equals("/at")) {
+                eventIndex = i + 1;
+                for(int j = eventIndex ; j < words.length ; j++) {
+                    timeAllocation = timeAllocation + " " + words[j];
+                }
+                break;
+            }
+            eventDescription = eventDescription + " " + words[i];
+        }
```
kaiwen98: Suggest to modify this code to avoid nested loop. You may want to look into Java String methods.

PR #42, opened 2021-08-28
[Deon Chung] iP

deonchung:


src/main/java/Duke.java
```diff
@@ -1,10 +1,106 @@
...
+
+    private static void addToDo(String inputCommand) {
+        System.out.println(HORIZONTAL_LINE);
+        String taskToDo = inputCommand.substring(5);
+        tasks.add(new ToDo(taskToDo));
+        GotItMessage();
+        System.out.println(HORIZONTAL_LINE);
+    }
+
+    private static void GotItMessage() {
```
LilyDaytoy:
```suggestion
    private static void gotItMessage() {
```

LilyDaytoy: is it better here to follow the naming standard that using lowercase first "gotItMessage()"?

src/main/java/Duke.java
```diff
@@ -1,10 +1,106 @@
...
+        System.out.println("Got it. I've added this task:");
+        System.out.println(tasks.get(tasks.size() - 1).toString());
+        System.out.println("Now you have " + tasks.size() + " tasks in the list.");
+    }
+
+    private static void MarkAsDone(String inputCommand) {
```
LilyDaytoy:
```suggestion
    private static void markAsDone(String inputCommand) {
```
is it better here to follow the naming standard that using lowercase first "markAsDone()"?

src/main/java/Duke.java
```diff
@@ -1,10 +1,106 @@
+import java.util.Scanner;
+import java.util.ArrayList;
+
 public class Duke {
+
+    public static final String HORIZONTAL_LINE = "_____________________________________________";
```
LilyDaytoy: I like the way how he uses "constant" to represent the symbol

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,220 @@
...
+public class Duke {
+
+    public static final String HORIZONTAL_LINE = "_____________________________________________";
+    public static final String FILEPATH = "C:\\Users\\GMSF\\Documents\\ip\\Duke.txt";
+    private static ArrayList<Task> tasks = new ArrayList<Task>();
+
+    public static void main(String[] args) {
+        loadData();
+        printHello();
+        task();
```
OngDeZhi: May want to give it a better name so it is obvious to readers as to what the method task is about.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,220 @@
...
+        for (Task item : tasks) {
+            file.append(item.toSave() + System.lineSeparator());
+        }
+        file.close();
+    }
+
+    public static void loadData() {
+        try {
+            File file = new File(FILEPATH);
+            Scanner s = new Scanner(file);
```
OngDeZhi: Could probably use a better variable name like scanner or something.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,220 @@
...
+                String taskDescription = descriptionParts[1];
+                Boolean isCompleted = Boolean.parseBoolean(descriptionParts[2]);
+                try {
+                    Task task;
+                    switch (typeOfTask) {
+                    case "todo":
+                        task = new ToDo(taskDescription);
+                        break;
+                    case "deadline":
+                        task = new Deadline(taskDescription, descriptionParts[3]);
```
OngDeZhi: Can create a variable to store descriptionParts[3] so the readers know what it is.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,220 @@
...
+                    addDeadline(inputCommand);
+                } else if (inputCommand.contains("event")) {
+                    addEvent(inputCommand);
+                } else if (inputCommand.contains("delete")){
+                    deleteTask(inputCommand);
+                } else if(inputCommand.contains("bye")){
+                    try {
+                        saveData();
+                    } catch (IOException e){
+                    }
```
OngDeZhi: Avoid empty catch block, leave comment if you intend it to be.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,220 @@
...
+    private static void deleteTask(String inputCommand) {
+        System.out.println(HORIZONTAL_LINE);
+        try {
+            int position = inputCommand.indexOf(" ");
+            int taskToDelete = Integer.parseInt(inputCommand.trim().substring(position + 1)) - 1;
+            Task removeTask = tasks.get(taskToDelete);
+            tasks.remove(taskToDelete);
+            System.out.println("Noted! I've remove this task:");
+            System.out.println(removeTask.toString());
+            System.out.println("Now you have " + tasks.size() + " tasks in the list.");
```
OngDeZhi: Can consider using line spacing to group related segments of code.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,220 @@
...
+        System.out.println(HORIZONTAL_LINE);
+    }
+
+    private static void printBye() {
+        System.out.println(HORIZONTAL_LINE);
+        System.out.println("Bye. Hope to see you again soon!");
+        System.out.println(HORIZONTAL_LINE);
+    }
+
+    public static void task() {
```
OngDeZhi: May want to consider applying SLAP in this method.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,220 @@
...
+            String taskToDo = inputCommand.substring(5);
+            tasks.add(new ToDo(taskToDo));
+            gotItMessage();
+        } catch (StringIndexOutOfBoundsException e) {
+            System.out.println("☹ OOPS!!! The description of a todo cannot be empty.");
+        }
+        System.out.println(HORIZONTAL_LINE);
+    }
+
+    private static void gotItMessage() {
```
OngDeZhi: Should use a verb here, like print at the front.

PR #43, opened 2021-08-28
[Sim Tian Boon] iP

Kair0s3:


src/main/java/Duke.java
```diff
@@ -1,10 +1,252 @@
...
+import java.util.Arrays;
+
 public class Duke {
+
+    /**
+     * Decorative prefixes to format program's output more neatly/cleaner.
+     */
+    private static final String CONSOLE_LINE_PREFIX = "____________________________________________________________";
+    private static final String SPACE_PREFIX = " ";
+    private static final String LINE_BREAK = "\n";
```
peng-217: Good for using all upper cases

peng-217: for naming constant strings

src/main/java/Duke.java
```diff
@@ -1,10 +1,252 @@
...
+    // Prefixes meant for processing data parameters for Task
+    private static final String BY_WHEN_PREFIX = "/by";
+    private static final String AT_WHEN_PREFIX = "/at";
+    private static final String TASK_TODO_PREFIX = "Todo";
+    private static final String TASK_DEADLINE_PREFIX = "Deadline";
+    private static final String TASK_EVENT_PREFIX = "Event";
+
+    /**
+     * Initializing Scanner variable to allow for reading User Inputs.
+     */
```
peng-217: All the comments clearly representing the functions of each method. And the naming is good since all methods' names start with verbs

src/main/java/Event.java
```diff
@@ -0,0 +1,13 @@
...
+    public Event (String taskName, String atWhen) {
+        super(taskName);
+        this.atWhen = atWhen;
+    }
+
+    @Override
+    public String toString() {
+        return "[E]" + super.toString() + " (At -> " + atWhen + ")";
+    }
+}
```
peng-217: Maybe can add some comments to illustrate the functions of methods

peng-217 (commented):
Overall, your code is very nice as it strictly follows the coding standard and comments are also very clear.

src/main/java/Duke.java
```diff
@@ -1,10 +1,392 @@
...
+    private static final Scanner SC = new Scanner(System.in);
+
+    /**
+     * Maximum number of Tasks for a user.
+     */
+    private static final String COMMAND_DELETE = "Delete";
```
kwokyto: Comment is not relevant to the variable.

src/main/java/Duke.java
```diff
@@ -1,10 +1,392 @@
...
+            System.out.println(SPACE_PREFIX + "EEEEEOOOOOO~ ALL RIGHT~ Oops was jamming away in my virtual garage, here's your PLAN/S...");
+            for (int i = 0; i < taskCounter; i++) {
+                System.out.println(SPACE_PREFIX + (i + 1) + "." + SPACE_PREFIX + tasks.get(i));
+            }
+            System.out.println(CONSOLE_LINE_PREFIX);
+        }
+    }
+
+    /**
+     * Gets the user input from I/O
```
kwokyto: Perhaps you can use "Returns the ..." instead.

src/main/java/Duke.java
```diff
@@ -1,10 +1,392 @@
...
+            String actualTaskName = unprocessedTaskName.replace(AT_WHEN_PREFIX, "").replace(atWhen, "");
+            actualTaskName = actualTaskName.trim();
+            return new Event(actualTaskName, atWhen);
+        } catch (ArrayIndexOutOfBoundsException arrError) {
+            throw new DukeException("TOBEREPLACED WITH EVENT_EMPTY_MESSAGE");
+        }
+    }
+
+    private static Task createSavedTask(String fileLine) {
+        String[] splitByDelimiter = fileLine.split("\\|\\|");
```
kwokyto: Perhaps this variable could have a different name that better explains what it is.

src/main/java/Duke.java
```diff
@@ -1,10 +1,392 @@
...
+        Task savedTask;
+        switch (splitByDelimiter[0]) {
+        case "D":
+            savedTask = new Deadline(splitByDelimiter[2],
```
kwokyto: This seems like a complicated expression, perhaps you can split it into separate lines?

src/main/java/Duke.java
```diff
@@ -1,10 +1,392 @@
...
+        default:
+            savedTask = new Todo(splitByDelimiter[2], Boolean.valueOf(splitByDelimiter[1]));
+        }
+        return savedTask;
+    }
+
+    /**
+     * Adds a new Task to list of Tasks.
+     *
+     * @param userInput Contains the type of task and relevant details
```
kwokyto: I think the input description could be refined to show an object rather than an action.

src/main/java/Duke.java
```diff
@@ -1,10 +1,392 @@
...
+                char taskIdentifier = tasks.get(i).toString().charAt(1);
+                fw.write(taskIdentifier + "||" + tasks.get(i).isDone() + "||" + tasks.get(i).getTaskName());
+                if (taskIdentifier == 'D') {
+                    fw.write("||" + ((Deadline) tasks.get(i)).getByWhen());
+                } else if (taskIdentifier == 'E') {
+                    fw.write("||" + ((Event) tasks.get(i)).getAtWhen());
+                }
+                fw.write(LINE_BREAK);
+            }
+            fw.close();
```
kwokyto: This seems like deep nesting, is there another way to represent this?

src/main/java/Duke.java
```diff
@@ -1,10 +1,392 @@
...
+                Scanner fileScanner = new Scanner(savedFile);
+                while (fileScanner.hasNext()) {
+                    String fileLine = fileScanner.nextLine();
+                    Task savedTask = createSavedTask(fileLine);
+                    tasks.add(savedTask);
+                    taskCounter++;
+                }
+            } catch (FileNotFoundException e) {
+                e.printStackTrace();
+            }
```
kwokyto: This seems like arrowhead style code, perhaps there is a way to make the happy path more prominent?

src/main/java/Duke.java
```diff
@@ -1,10 +1,392 @@
...
+                    if (userParams[0].equalsIgnoreCase(COMMAND_DONE)) {
+                        int index = Integer.parseInt(userParams[1]);
+                        markTaskAsDone(index - 1);
+                    } else if (userParams[0].equalsIgnoreCase(COMMAND_DELETE)) {
+                        int index = Integer.parseInt(userParams[1]);
+                        deleteTask(index - 1);
+                    } else {
+                        addToTasks(userInput);
+                    }
+                }
```
kwokyto: This seems like arrowhead style code, is there a way to make the happy path prominent?

src/main/java/Duke.java
Kair0s3: I have made some changes, but I'm not very sure if it solves the deep nesting issue.

PR #45, opened 2021-08-28
[Aloysius Lee] iP

arraysius:


src/main/java/Duke.java
```diff
@@ -1,10 +1,271 @@
...
+     */
+    private static void blockPrint(String[] sentences) {
+        String printMessage = LINE + String.join("\n", sentences) + "\n" + LINE;
+        System.out.println(printMessage);
+    }
+
+    /**
+     * Print help message with valid commands.
+     */
+    private static void printHelp() {
```
itsleeqian, line 38: all in all, variable and function names are all good. Good job!

itsleeqian: All in all, good naming of functions and variables. Functions are in the form of verbs. Good job!

src/main/java/Duke.java
```diff
@@ -34,17 +111,51 @@ private static void listTasks() {
...
+            addDeadline(splitUserInput);
+            break;
+        case COMMAND_EVENT:
+            addEvent(splitUserInput);
+            break;
+        default:
+            printHelp();
+            break;
+        }
     }
```
nguyin: I like how you used case instead of if-else statements, then you will avoid the deep nesting :)

src/main/java/Duke.java
```diff
@@ -34,17 +111,51 @@ private static void listTasks() {
...
+            taskListMessage[i + 1] = (i + 1) + ". " + task.toString();
         }
 
         blockPrint(taskListMessage);
     }
 
     private static void markTaskAsDone(int taskIndex) {
+        if (taskManager.getTotalTasks() == 0) {
+            blockPrint(new String[]{"The list is currently empty. Add a task first."});
+            return;
```
nguyin: Nice job handling the error case before dealing with the actual conditions

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,53 @@
...
+
+public class Task {
+
+    private String description;
+    private boolean isDone;
+    /** Type of task (Todo [T], Deadline [D], Event [E]). */
+    private String type = " ";
+
+    /**
+     * Create a new Task object.
```
kwokyto: Would be better if it is "Creates"

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,53 @@
+package duke.task;
+
+public class Task {
+
+    private String description;
+    private boolean isDone;
+    /** Type of task (Todo [T], Deadline [D], Event [E]). */
+    private String type = " ";
```
kwokyto: Is this a magic literal? How is this value used in the rest of your code?

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,53 @@
...
+        return description;
+    }
+
+    /**
+     * Returns "X" if the task is done and a whitespace otherwise.
+     *
+     * @return Completion status icon.
+     */
+    public String getStatusIcon() {
+        // Mark done task with X
```
kwokyto: I think this comment repeats information stated in the Javadoc comments

src/main/java/duke/task/TaskManager.java
```diff
@@ -0,0 +1,67 @@
...
+     * Create task manager.
+     */
+    public TaskManager() {
+    }
+
+    /**
+     * Checks if task index is within task list size.
+     *
+     * @param taskIndex Index of task
+     * @return Returns true if index is within task list size, otherwise false.
```
kwokyto: Perhaps the comment could be "@return boolean indicating if index is within task list size"

src/main/java/duke/task/TaskManager.java
```diff
@@ -0,0 +1,67 @@
...
+     */
+    private boolean hasTaskIndex(int taskIndex) {
+        return (taskIndex >= 0) && (taskIndex < getTotalTasks());
```
kwokyto: This seems like a complicated expression, perhaps it would be better to split it into different lines?

src/main/java/duke/task/TaskManager.java
```diff
@@ -0,0 +1,67 @@
...
+    /**
+     * Create task manager.
+     */
+    public TaskManager() {
+    }
+
+    /**
+     * Checks if task index is within task list size.
+     *
+     * @param taskIndex Index of task
```
kwokyto: I think you missed a fullstop at the end of the comment

src/main/java/duke/task/TaskManager.java
```diff
@@ -0,0 +1,67 @@
...
+    private boolean hasTaskIndex(int taskIndex) {
+        return (taskIndex >= 0) && (taskIndex < getTotalTasks());
+    }
+
+    public int getTotalTasks() {
+        return taskList.size();
+    }
+
+    /**
+     * Get the task with the specified index from the task list.
```
kwokyto: Perhaps it is better to say "Returns the task"

src/main/java/duke/task/TaskManager.java
```diff
@@ -0,0 +1,67 @@
...
+     *
+     * @param taskIndex Index of the task in the task list.
+     * @return Task.
+     */
+    public Task getTask(int taskIndex) {
+        return this.taskList.get(taskIndex);
+    }
+
+    /**
+     * Add task into task list.
```
kwokyto: Perhaps it is better to say "Adds"

PR #50, opened 2021-08-28
[Jon Lim] iP

jon-the-melon:


src/main/java/Duke.java
```diff
@@ -1,10 +1,178 @@
...
+    private static final String HELP_MESSAGE = " Here are the commands for the things I can do:\n" +
+            "    1. todo [task] = adds a Todo task\n" +
+            "    2. deadline [task] /[deadline] = adds a Deadline task\n" +
+            "    3. event [task] /[event range] = adds an Event task\n" +
+            "    4. done [taskIndex] = marks the inputted task as done\n" +
+            "    5. list = lists out all current tasks with their taskIndex\n" +
+            "    6. echo = turn into a huge cave and echo your inputs back to you!!\n" +
+            "    7. quit (only in echo mode) = turn back to normal and stop echoing\n" +
+            "    8. genshin = begin the genshin helper\n" +
+            "    9. bye = shuts me down... ;-;\n";
```
jyxhazcake: This string is a little wordy. Perhaps you can consider a method which grabs each of the tasks' helper String, and then displays all for the user, This will help to reduce the need for a long String like this and makes it easier for you to edit this Help Message String as well.

src/main/java/Duke.java
```diff
@@ -1,10 +1,178 @@
...
+            genshinHelper();
+        }
+        //Exit
+        else if (input.equalsIgnoreCase("bye")) {
+            System.out.println(LINES + BYE_MESSAGE + LINES);
+            System.exit(0);
+        }
+        else {
+            System.out.println(LINES + INVALID_COMMAND + LINES);
+        }
```
jyxhazcake: Lengthy method. Do try and extract this method out into more methods to achieve the Single Level of Abstraction Principle.

src/main/java/Duke.java
```diff
@@ -1,10 +1,178 @@
...
+        else {
+            int slashIndex = input.indexOf("/");
+            String task = input.substring(6, (slashIndex-1));
+            String timeRange = input.substring(slashIndex+4);
+            tasks[taskCount] = new Event(task, timeRange);
+        }
+        System.out.println(LINES + ADD_MESSAGE + tasks[taskCount]);
+        taskCount += 1;
+        System.out.println(" Now you have " + taskCount + " tasks in the list.\n" + LINES);
+        return taskCount;
```
jyxhazcake: Good use of if and else ifs to avoid deep nesting, but perhaps you could extract this method down even further as it is a little lengthy.

src/main/java/Duke.java
```diff
@@ -1,10 +1,178 @@
...
+        String high = sc.nextLine();
+        totalMats += Integer.parseInt(high)*9;
+        System.out.print(" How many Mid Tier: ");
+        String mid = sc.nextLine();
+        totalMats += Integer.parseInt(mid)*3;
+        System.out.print(" How many Low Tier: ");
+        String low = sc.nextLine();
+        totalMats += Integer.parseInt(low);
+        System.out.println(" Total low tier mats required: " + totalMats + "\n" + LINES);
+    }
```
jyxhazcake: The naming of this method is a little confusing, not quite sure but what this method does. Perhaps you can add comments and replace the use of magic numbers like "27", "9", "3" with properly-named constants.

jyxhazcake (commented):
Reviewing for code quality. Overall, a valiant effort. I see the effort in trying to extract methods and keeping the code plain and simple, but some methods and constants are still lengthy and a little confusing. Good job nonetheless and keep up the good work.

src/main/java/Duke.java
```diff
@@ -1,10 +1,178 @@
+import java.util.Scanner;
```
STAung07: I like how specific imports were used for ease of reference.

src/main/java/Duke.java
STAung07: Could consider providing documentation for what the function does to allow for ease of understanding for future collaboration.
Could consider giving meaningful names for each of the numbers 3, 9 and 27

src/main/java/Duke.java
```diff
@@ -1,10 +1,178 @@
...
+    public static void markDone(String input, Task[] tasks, int taskCount) {
+        //isolate 'x' from 'done x', where x is a number
+        int index = Integer.parseInt(input.substring(5));
+        if (index > taskCount) {
+            System.out.println(LINES + NO_TASK_MESSAGE + LINES);
+        }
+        else {
+            tasks[index - 1].markAsDone();
+            System.out.print(LINES + TASK_COMPLETE_MESSAGE + tasks[index-1] + "\n" + LINES);
+        }
```
STAung07: if-else format should be in the same standardized format as line 49-55. same for line 82-97

src/main/java/Event.java
```diff
@@ -0,0 +1,21 @@
+public class Event extends Task {
+    protected String from;
```
STAung07: Could consider naming it "at" for ease of understanding that event is on the day itself

src/main/java/JimException.java
```diff
@@ -0,0 +1,2 @@
+public class JimException extends Exception {
```
Samuel787: This seems to be an empty class. Perhaps you may want to implement a custom message for this exception.

src/main/java/Duke.java
```diff
@@ -1,10 +1,411 @@
...
+            " /by [deadline]'!" + System.lineSeparator();
+    private static final String MISSING_IN_EVENT_MESSAGE = "Please format your input as 'event [task]" +
+            " /at [time range]'!" + System.lineSeparator();
+    private static final String MISSING_BY_MESSAGE = "Please add '/by' in between your task and deadline!" +
+            System.lineSeparator();
+    private static final String MISSING_AT_MESSAGE = "Please add '/at' in between your task and time range!" +
+            System.lineSeparator();
+    private static final String INVALID_COMMAND = "I don't quite understand :/" + System.lineSeparator();
+    private static final String HELP_MESSAGE = " Here are the commands for the things I can do:" +
+            System.lineSeparator() +
```
Samuel787: There are too many constants in one file. You may consider either:
1. grouping related constants together and separating the group of constants with spacing
2. Extract out the logic in this class into different classes by making use of OOP concepts

src/main/java/Duke.java
```diff
@@ -1,10 +1,411 @@
...
+                echoState = false;
+            } else {
+                String output = LINES + " " + s  + System.lineSeparator() + LINES;
+                System.out.println(output);
+            }
+        }
+    }
+
+    //list function
+    public static void list() {
```
Samuel787: You may want to consider a more descriptive method name such as listTasks to provide more idea as to what exactly it lists.

src/main/java/Duke.java
```diff
@@ -1,10 +1,411 @@
...
+                System.out.println(i + "." + tasks.get(i - 1));
+            }
+            System.out.print(LINES);
+        }
+    }
+
+    //Removes a specified task
+    public static void removeTask(String input) {
+        try {
+            int index = Integer.parseInt(input.substring(7));
```
Samuel787: I'm not too sure where this 7 comes from. Seems like a magic number to me.

src/main/java/Duke.java
```diff
@@ -1,10 +1,411 @@
...
+                String dueDate = input.substring(slashIndex + 3).trim();
+                tasks.add(new Deadline(task, dueDate));
+                int taskIndex = tasks.size() - 1;
+                printTaskAcknowledgement(tasks.get(taskIndex));
+            }
+        } catch (StringIndexOutOfBoundsException e) {
+            System.out.println(LINES + MISSING_IN_DEADLINE_MESSAGE + LINES);
+        }
+    }
+    private static void addEvent(String input) {
```
Samuel787: You may want to add spacing between the end of addDeadline and start of addEvent.

src/main/java/Duke.java
```diff
@@ -1,10 +1,411 @@
...
+            fw.write("");
+            fw.close();
+        } catch (IOException e) {
+            System.out.println(LINES + NO_FILE_MESSAGE + LINES);
+        }
+    }
+
+    public static void updateDatabase() {
+        //update after mark done.
+        //clear file then append from list
```
Samuel787: You may want to move the comments outside of the method (just before the start of the method) and just describe at a high level what the method does if you wish to. Do not have to include implementation details in the comments unless the logic is extremely complex.

src/main/java/Duke.java
```diff
@@ -1,10 +1,411 @@
...
+        clearDatabase();
+        for (Task task: tasks) {
+            String isDone = task.isDone? "1" : "0";
```
Samuel787: "isDone" sounds like a boolean variable name. You may want to use a variable name such as completionStatus.

src/main/java/Duke.java
```diff
@@ -1,10 +1,411 @@
...
+                }
+                break;
+            default:
+                throw new JimException();
+            }
+        }
+        s.close();
+    }
+
+    //Initialises the list
```
Samuel787: This method seems to be initializing the program and not just Jim.

src/main/java/Duke.java
```diff
@@ -1,10 +1,411 @@
...
+        } catch (JimException e) {
+            System.out.println(LINES + "File is corrupted at Line " + (tasks.size()+1) + "!" +
+                    System.lineSeparator() + LINES);
+        } catch (FileNotFoundException e) {
+            System.out.println(LINES + NO_FILE_MESSAGE + LINES);
+        }
+    }
+
+    public static void folderChecker(File folder) {
+        boolean result = folder.mkdir();
```
Samuel787: You may want to use a name such as "isFolderCreated" that's more appropriate for boolean variables.

src/main/java/Duke.java
```diff
@@ -1,10 +1,411 @@
...
+        if (result) {
+            System.out.println("Folder not present...creating new folder!" + System.lineSeparator() + LINES);
+        } else {
+            System.out.println("Folder located!" + System.lineSeparator() + LINES);
+        }
+    }
+
+    public static void databaseChecker(File file) {
+        try {
+            boolean result = file.createNewFile();
```
Samuel787: Once again, you may want to use a name such as "isFolderCreated" that's more appropriate for boolean variables.

Samuel787 (commented):
Looks good for the most part. Duke.java class seems a bit too long. You may want to think if you can use OOP further to extract the logic out further into different classes.

PR #54, opened 2021-08-28
[Wei Xuan] iP

WeiXuanYap:


src/main/java/Duke.java
```diff
@@ -1,10 +1,119 @@
...
+    private static final Scanner in = new Scanner(System.in);
+
+    /**
+     *
+     * List of all Tasks
+     */
+    private static Task[] allTasks;
+
+    private static String input;
+    private static String logo = getLogo();
```
hozhenhong99: can declare the logo as a final variable since it does not change

src/main/java/Duke.java
```diff
@@ -1,10 +1,119 @@
...
+            String deadline = userInput.substring(deadlineStartIndex);
+            Deadline newDeadline = new Deadline(taskName, deadline);
+            allTasks[inputCount] = newDeadline;
+            inputCount += 1;
+            taskMessage(inputCount, newDeadline);
+        }
+        else if(userInput.startsWith("event ") && userInput.contains("/at")) {
+            int taskEndIndex = userInput.indexOf("/at") - 1; //to account for spacing before "/by"
+            String taskName = userInput.substring(6, taskEndIndex);
+            int eventTimeStartIndex = taskEndIndex + 5;
```
hozhenhong99: can consider storing the numbers as static final variables so we know how they originate

src/main/java/Duke.java
```diff
@@ -1,10 +1,119 @@
...
+            System.exit(0);
+        }
+        else if(userInput.equals("list")) {
+            for(int i = 0; i < inputCount; i++){
+                //System.out.println((i+1) + ". [" + inputs[i].getStatusIcon() + "] " +
+                System.out.println((i+1) + "." + allTasks[i]);
+            }
+        }
+        //assumes user will follow format "done x"
+        else if(userInput.startsWith("done ")) {
```
hozhenhong99: is there a case where user tries to mark a task not in the list as done? e.g. list has 2 items and user does 'done 5'

src/main/java/Duke.java
```diff
@@ -1,10 +1,119 @@
...
+
+    /**
+     *
+     * List of all Tasks
+     */
+    private static Task[] allTasks;
+
+    private static String input;
+    private static String logo = getLogo();
+    private static String line = getLine();
```
hozhenhong99: This variable can be declared as final as it does not change

src/main/java/Deadline.java
```diff
@@ -0,0 +1,21 @@
+public class Deadline extends Task {
+
+    protected String by;
```
DJ-Tan: Naming - Consider using a more intuitive name such as "dueDate" instead of "by", since "by" used in certain context may not have a date meaning attached to it

src/main/java/Duke.java
DJ-Tan: Agree with Zhen Hong as well. Consider storing logo and line as a final String variable. If the logo takes up a lot of space, consider defining a new Java class and putting all these constants in there.

src/main/java/Duke.java
```diff
@@ -1,10 +1,119 @@
...
+    }
+
+    private static void taskQuery(String userInput) {
+        if(userInput.equals("bye")) {
+            System.out.println("Bye, see you!\n" + line);
+            System.exit(0);
+        }
+        else if(userInput.equals("list")) {
+            for(int i = 0; i < inputCount; i++){
+                //System.out.println((i+1) + ". [" + inputs[i].getStatusIcon() + "] " +
```
DJ-Tan: Remove commented out code from the submission

src/main/java/Duke.java
```diff
@@ -1,10 +1,119 @@
...
+                System.out.println((i+1) + "." + allTasks[i]);
```
DJ-Tan: Leave a space - "i+1" should be "i + 1"

src/main/java/Duke.java
DJ-Tan: Agree with Zhen Hong. Do no use magic numbers. For example in words[1] instead of using 1, use a constant such as TASK_NAME or DATE to describe it

src/main/java/Duke.java
```diff
@@ -1,10 +1,119 @@
...
+    }
+
+    private static void initTaskList() {
+        allTasks = new Task[100];
+    }
+
+    private static void taskMessage(int inputCount, Task task) {
+        System.out.println("Successfully added. Here's the added task: ");
+        System.out.println(task);
+        System.out.println("There " + ((inputCount == 1)? "is " : "are ") + inputCount + " task" + ((inputCount == 1)? "" : "s") + " in the list now.");
```
DJ-Tan: Having two if else conditions described with ternary operators in a single line makes it hard for another user to read

src/main/java/Duke.java
```diff
@@ -1,10 +1,119 @@
...
+    }
+
+    private static void welcomeMessage(String logo, String line) {
+        System.out.println(line + logo +
+                " Hello! I'm Duke\n" +
+                " What's up? :p\n" + line);
+    }
```
DJ-Tan: According to this page: https://se-education.org/guides/conventions/java/basic.html
The "+" goes on the next line.
So we have... 
System.out.println(line + logo
        + "Hello! I'm Duke\n"
        + "What's up? :p\n" + line);

src/main/java/Event.java
```diff
@@ -0,0 +1,20 @@
+public class Event extends Task{
+    protected String on;
```
DJ-Tan: "eventDate" may be a better naming than "on"

src/main/java/ToDo.java
```diff
@@ -0,0 +1,21 @@
+public class ToDo extends Task {
+    protected boolean isDone;
+
+    public boolean isDone() {
+        return isDone;
+    }
+
+    public void setDone(boolean done) {
+        isDone = done;
+    }
```
DJ-Tan: If the statement is always true, can set isDone = true instead

PR #57, opened 2021-08-28
[An Sheng] iP

AnShengLee:


src/main/java/UserInterface.java
```diff
@@ -2,96 +2,116 @@
...
+                break;
+            }
+            showInvalidIndex();
+            break;
+            
+        default:
+            showInvalidCommand();
+            break;
+        }
     }
```
jerrelllzw: Is there a need for spaces between cases?

src/main/java/UserInterface.java
```diff
@@ -2,96 +2,116 @@
...
-            if (!TaskManager.checkCorrectIndex(taskIndex)) {
-                showInvalidIndex();
-                return Command.CONTINUE;
-            }
-            TaskManager.setDone(taskIndex);
-            showItemRemoved(taskIndex);
-            return Command.CONTINUE;
+    private static void printLine() {
+        System.out.println(HORIZONTAL_LINE);
+    }
```
jerrelllzw: Good job on the "{}" brackets for a single line code.

src/main/java/UserInterface.java
```diff
@@ -2,96 +2,116 @@
 
 public class UserInterface {
     private static final String HORIZONTAL_LINE = "____________________________________________________________";
```
jerrelllzw: Appreciate that all the constants are declared.

src/main/java/Command.java
```diff
@@ -4,5 +4,6 @@
     ADD_TODO,
```
jerrelllzw: Good job on using underscores to separate words for constants.

src/main/java/UserInterface.java
AnShengLee: You are right I think there is no need for a space between cases. Thank you.

src/main/java/duke/UserInterface.java
```diff
@@ -0,0 +1,232 @@
+package duke;
+
+import duke.command.Command;
+
```
ongweisheng, line 7: Consider separating UserInterface class into Parser and UI to for further abstraction. Currently it is fine regardless

src/main/java/duke/TaskManager.java
```diff
@@ -0,0 +1,142 @@
...
+     * If the Task index given is valid.
+     * 
+     * @param taskIndex The index of Task we are deleting.
+     * @return The deleted Task.
+     * @throws DukeInvalidTaskIndexException If the index given exceeds the range of current number of Task.
+     */
+    public static Task delete(int taskIndex) throws DukeInvalidTaskIndexException {
+        if (checkCorrectIndex(taskIndex)) {
+            Task deleting = tasks.get(taskIndex - 1);
+            tasks.remove(taskIndex - 1);
```
ongweisheng: Consider offsetting taskIndex with -1 beforehand somewhere instead of indicating it like this in the code

src/main/java/duke/UserInterface.java
```diff
@@ -0,0 +1,232 @@
...
+            return Command.ADD_TODO;
+        } else if ("deadline".equals(userInputs[USER_COMMAND_INDEX])) {
+            return Command.ADD_DEADLINE;
+        } else if ("event".equals(userInputs[USER_COMMAND_INDEX])) {
+            return Command.ADD_EVENT;
+        } else if ("done".equals(userInputs[USER_COMMAND_INDEX])) {
+            return Command.DONE;
+        } else if ("delete".equals(userInputs[USER_COMMAND_INDEX])) {
+            return Command.DELETE;
+        }
```
ongweisheng: Consider changing the different types of command e.g. todo, deadline into constants, you may need to use this constants elsewhere as well

src/main/java/duke/DukeBlankDescriptionsException.java
```diff
@@ -0,0 +1,4 @@
+package duke;
+
+public class DukeBlankDescriptionsException extends Exception {
+}
```
ongweisheng: Consider creating packages for your different exceptions. Also applies to your other classes as well that may fit under a category

ongweisheng (commented):
Generally job well done, try and see if you can abstract more code into a more specific class. Also, try and include classes of the same functionality into a package e.g. exceptions, deadline, todo etc.

src/main/java/duke/TaskManager.java
AnShengLee: Thank you. Will change it in future commits.

src/main/java/duke/UserInterface.java
AnShengLee: Thank you. Will change it in future commits.

src/main/java/duke/DukeBlankDescriptionsException.java
AnShengLee: Thank you. Will change it in future commits.

src/main/java/duke/UserInterface.java
AnShengLee: Thank you. Might add the Parser class if time permits.

PR #66, opened 2021-08-29
[Pranav Venkatram] iP

giterator:


src/main/java/Todo.java
```diff
@@ -0,0 +1,13 @@
+public class Todo extends Task {
+    public static final char TASK_TYPE = 'T';
```
richwill28, line 5: Any reason of doing this? This line seems redundant to me.
Consider the following alternative instead when returning the function on [line 11](https://github.com/giterator/ip/blob/50c6f3df99cbeb630576a263c1b985d1aa2a4e22/src/main/java/Todo.java#L11).
```java
return "[T]" + "[" + completeCharacter + "] " + description;
```

I also noticed the same implementations on the `Deadline` and `Event` class as well.

src/main/java/Response.java
```diff
@@ -0,0 +1,53 @@
+public class Response {
```
richwill28: Consider adding a `JavaDoc` comment to explain the behavior of the public class, since it will be used by others.

src/main/java/Storage.java
```diff
@@ -0,0 +1,40 @@
...
+        }
+
+        System.out.println(Response.getLine());
+    }
+
+    public static void markComplete(int completedTask) {
+        inputTasks[completedTask - 1].markComplete();
+        Response.echo("Nice! I've marked this task as done: \n" + inputTasks[completedTask - 1]);
+    }
+}
```
richwill28: File does not end with a newline. Consider adding one to conform with the git file convention.

richwill28 (commented):
Overall, looks good to merge. I like that you created two new classes `Response` and `Storage` to help with the separation of concerns.

src/main/java/Deadline.java
```diff
@@ -0,0 +1,15 @@
+public class Deadline extends Task {
+    public static final char TASK_TYPE = 'D';
+    private final String BY;
```
Poopies99, line 3: variable names must in camelCase which in this case "private final String by;" would be appropriate

src/main/java/Response.java
```diff
@@ -0,0 +1,53 @@
...
+
+    }
+
+    public static void parseInput(String input) {
+        if (input.startsWith("done ")) {
+            String[] inputParts = input.split(" ");
+            int completedTask = Integer.parseInt(inputParts[1]);
+            Storage.markComplete(completedTask);
+            return;
+        }
```
Poopies99: Code is unnecessarily elongated, consider adding the above if statement within the switch statement below

src/main/java/Deadline.java
```diff
@@ -0,0 +1,15 @@
+public class Deadline extends Task {
+    public static final char TASK_TYPE = 'D';
```
Poopies99, line 5: Consider removing the above variable and instead incorporating it into the toString() method 

example: 
return "[D]" + ...

src/main/java/Task.java
```diff
@@ -0,0 +1,13 @@
+public class Task {
+    protected String description;
+    protected boolean isComplete;
```
Poopies99, line 6: Good job on naming the variables in the class with camelCase :)

Poopies99 (commented):
Good job, some minor changes will do :)

src/main/java/Deadline.java
Poopies99: Will need to comply with the Java Coding Standard as variables in class needs to be in camelCase

Consider changing it to "by" instead

src/main/java/Storage.java
```diff
@@ -0,0 +1,40 @@
+public class Storage {
+    private static final Task[] inputTasks = new Task[100];
+    private static int inputTasksSize = 0;
```
Poopies99: Good job on naming the variable names with camelCase :)

Poopies99 (commented):
LGTM

src/main/java/Storage.java
```diff
@@ -0,0 +1,40 @@
...
+
+    public static void storeInput(String input) {
+        Task newTask;
+        if (input.startsWith("todo")) {
+            newTask = new Todo(input.substring(5));
+        } else if (input.startsWith("deadline")) {
+            String[] inputSlices = input.substring(9).split("/");
```
KaiserHuang88: Can try removing on explaining (or removing) the magic numbers (9 and 5) here.

src/main/java/Response.java
```diff
@@ -0,0 +1,53 @@
+public class Response {
+    private static final String LINE = "____________________________________________________________";
```
KaiserHuang88: I believe using LINE like this doesn't comply with the coding standard

src/main/java/Duke.java
```diff
@@ -1,10 +1,17 @@
...
-        System.out.println("Hello from\n" + logo);
+        Response.greet();
+
+        String input;
+        do {
+            Scanner scan = new Scanner(System.in);
+            input = scan.nextLine();
+
+            Response.parseInput(input);
+        } while (!input.equals("bye"));
```
KaiserHuang88: Can squeeze the obtaining of input into a single function to make the code look cleaner.

src/main/java/Response.java
```diff
@@ -0,0 +1,53 @@
...
+        switch (input) {
+        case "bye":
+            bye();
+            break;
+        case "list":
+            Storage.list();
+            break;
+        default:
+            Storage.storeInput(input);
+            break;
```
KaiserHuang88: Consider adding a comment here to explain what switch does in this case.

KaiserHuang88 (commented):
Well Done! Very few mistakes that can be spotted.

src/main/java/Deadline.java
okkhoy: There is nothing wrong in declaring a constant for specific keywords. I think this is fair enough.
Using `"[D]"` results in the code reader asking the question: why is it `"[D]"` and not any other letter?

src/main/java/Deadline.java
```diff
@@ -0,0 +1,41 @@
...
+     * @return The String representation of the Deadline.
+     */
+    @Override
+    public String toString() {
+        completeStatus = isComplete ? COMPLETE_CHARACTER : INCOMPLETE_CHARACTER;
+        return "[" + TASK_TYPE + "]" + "[" + completeStatus + "] " + description + " (by: " + BY + ")";
+    }
+
+    /**
+     * Returns a String object representing the Deadline in a format that can be used to read and write
```
okkhoy, line 33: Good use of JavaDoc comments. Furthermore, the format is correct too. 👍

src/main/java/DukeException.java
```diff
@@ -0,0 +1,2 @@
+public class DukeException extends Exception {
+}
```
okkhoy, line 2: Exploit the fact that you are creating a custom exceptions? e.g., you can consider adding a custom message to your custom exception.

src/main/java/Event.java
```diff
@@ -0,0 +1,41 @@
...
+ */
+public class Event extends Task {
+    public static final char TASK_TYPE = 'E';
+    private final String AT;
+
+    /**
+     * Constructor for the Event class.
+     *
+     * @param description Details of an Event object.
+     * @param at          Finishing time of the Event object.
```
okkhoy, line 14: 💯

src/main/java/Parser.java
```diff
@@ -0,0 +1,52 @@
...
+     * Parses the input command given by the user to determine which functionality to execute.
+     *
+     * @param input String entered by the user, denoting instructions for the program.
+     */
+    public static void parseInput(String input) {
+        String[] inputParts = input.split(" ");
+        String instruction = inputParts[0];
+
+        switch (instruction) {
+        case "done":
```
okkhoy: Hmmm... here you could consider using command word as constants instead of "magic literals"

src/main/java/Parser.java
```diff
@@ -0,0 +1,52 @@
...
+        }
+    }
+
+    /**
+     * Attempts the store feature on the given user input. If the input does not correspond to todo, deadline or event,
+     * a DukeException is caught. This signifies a wrong input command or an incomplete description of the given task.
+     *
+     * @param input Input command given by the user
+     */
+    private static void attemptStore(String input) {
```
okkhoy: Perhaps `store` is a better name. Now as a reader I get this question: Ok attempted to store, so where is exactly store happening?
Exception is a mechanism to say failed storing. Hope you get the drift?

src/main/java/Storage.java
```diff
@@ -0,0 +1,102 @@
...
+    /**
+     * Loads stored tasks from file.
+     *
+     * @param storedTasks File at STORED_PATH.
+     * @throws FileNotFoundException If the file at STORED_PATH does not exist.
+     */
+    private static void loadStoredTasks(File storedTasks) throws FileNotFoundException {
+        Scanner scanner = new Scanner(storedTasks);
+        try {
+            scanner.nextLine(); //Moves past prefix \n of file
```
okkhoy: redundant comment here.

src/main/java/Storage.java
```diff
@@ -0,0 +1,102 @@
...
+            File storedTasks = new File(STORAGE_PATH);
+            if (storedTasks.createNewFile()) {
+                System.out.println("File created: " + storedTasks.getName());
+            } else {
+                System.out.println("Loading from " + STORAGE_PATH);
+                loadStoredTasks(storedTasks);
+            }
+        } catch (FileNotFoundException e) {
+            System.out.println("An error in reading " + STORAGE_PATH + " occurred.");
+            System.exit(OK_EXIT_CODE);
```
okkhoy: Sure this is `OK_EXIT_CODE`?

okkhoy: shouldn't it be an error?

src/main/java/Task.java
okkhoy: Hmmm... @Poopies99 , it would be better to say "good job on naming the boolean variable sound like boolean"

src/main/java/TaskList.java
```diff
@@ -0,0 +1,122 @@
...
+
+        acknowledgeMessage = "Got it. I've added this task: \n  " + inputTasks.get(inputTasks.size() - 1) + "\n"
+                + "Now you have: " + inputTasks.size() + " tasks in the list";
+
+        Ui.echo(acknowledgeMessage);
+
+        if (shouldAppendToFile) {
+            Storage.appendTaskToFile(newTask);
+        }
+    }
```
okkhoy: See your comment: `Parses ... and stores ...`
So you can consider splitting the method to 2 different methods each handling its own functionality: parsing and storing.

src/main/java/Todo.java
okkhoy: @richwill28  I suppose this is OK? extracting a constant

okkhoy (commented):
Overall very neat code. Good job!
You may want to pay some attention to SLAP.

src/main/java/Todo.java
richwill28: @okkhoy Initially I thought it was redundant, but on second thought, I guess it is better than having magic constants which could caused readers to be confused. Thanks for the follow up, I was quite unsure about this part as well.

PR #76, opened 2021-08-29
[Hui En] iP

huien77:


src/main/java/Duke.java
```diff
@@ -1,10 +1,129 @@
...
+    private static final String LOGO =
+                    " _______  __   __  ______   _______  _______  _______  __   __  _______\n" +
+                    "|       ||  | |  ||      | |       ||       ||   _   ||  |_|  ||   _   |\n" +
+                    "|    ___||  | |  ||  _    ||    ___||_     _||  |_|  ||       ||  |_|  |\n" +
+                    "|   | __ |  |_|  || | |   ||   |___   |   |  |       ||       ||       |\n" +
+                    "|   ||  ||       || |_|   ||    ___|  |   |  |       ||       ||       |\n" +
+                    "|   |_| ||       ||       ||   |___   |   |  |   _   || ||_|| ||   _   |\n" +
+                    "|_______||_______||______| |_______|  |___|  |__| |__||_|   |_||__| |__|\n";
+    private static final String SEPARATOR = "zzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzz\n";
+    private static Task[] userInputs = new Task[100];
```
yuejunfeng0909: It's better to set a constant MAX_TASK = 100 instead of just put 100 here

src/main/java/Duke.java
```diff
@@ -1,10 +1,129 @@
...
+            }
+            line = in.nextLine();
+        }
+        endDuke();
+    }
+
+    /**
+     * Prints the ending messages before ending the program
+     */
+    public static void endDuke() {
```
yuejunfeng0909: This function is not doing what the name is suggesting, as it is not terminating the application

src/main/java/Duke.java
```diff
@@ -1,10 +1,129 @@
...
+    private static final String SEPARATOR = "zzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzz\n";
+    private static Task[] userInputs = new Task[100];
+    private static int userInputCount = 0;
+
+    /**
+     * Initiate the program and print the various messages
+     */
+    public static void initiateDuke() {
+        System.out.println(SEPARATOR + "Hi... from GUDETAMA... so sleepy\n" + LOGO);
+        System.out.println("Give me five more minutes..... What can I do for you?\n" + SEPARATOR);
```
naijie2108: You can consider using constants for messages for easier reading and maintaining of the code

src/main/java/Duke.java
```diff
@@ -1,10 +1,129 @@
...
+                addTodo(line);
+            } else if (line.startsWith("deadline")) {
+                addDeadline(line);
+            } else if (line.startsWith("event")) {
+                addEvent(line);
+            } else {
+                handleInvalid();
+            }
+            line = in.nextLine();
+        }
```
naijie2108: Good usage of single-level abstraction!

t-l-xin (commented):
Good job

good use of javadoc
good naming of function names 
good job at making line separator a constant to reuse in different files

src/main/java/Duke.java
```diff
@@ -1,10 +1,129 @@
...
+     * Marks a particular task in userInputs as done
+     * @param line the index of the task to be marked as done
+     */
+    public static void markDone(String line) {
+        int taskIndexDone = Integer.parseInt(line.split(" ")[1]) - 1;
+
+        userInputs[taskIndexDone].markAsDone();
+        System.out.println(SEPARATOR + "\n\tfinished this task... I need a break:\n\t\t" +
+                            userInputs[taskIndexDone]);
+        System.out.println(SEPARATOR);
```
naijie2108: I think the printing can be extracted to a new method (applying the SLAP method)

src/main/java/Duke.java
```diff
@@ -1,10 +1,129 @@
...
+        System.out.println(SEPARATOR + "\n\tadded: " + userInputs[userInputCount] + "\n" + SEPARATOR);
+        userInputCount++;
+    }
+
+    /**
+     * Adds a task with deadline into userInputs and increase the counter
+     * @param line the string command with the description and deadline of the task
+     */
+    public static void addDeadline(String line) {
+        String[] lineArray = line.substring(9).split("/by");
```
naijie2108: I think you can use a const instead of magic numbers here, as it isn't clear why 9 is used for the substring

src/main/java/Duke.java
```diff
@@ -2,58 +2,102 @@
...
+        System.out.println(SEPARATOR);
+    }
+
+    public static void markDone(String line) {
+        int taskIndexDone = Integer.parseInt(line.split(" ")[1]) - 1;
+
+        userInputs[taskIndexDone].markAsDone();
+        System.out.println(SEPARATOR + "\n\tfinished this task... I need a break:\n\t\t" +
+                            userInputs[taskIndexDone]);
+        System.out.println(SEPARATOR);
```
t-l-xin: SEPARATOR can be in the same print statement or since there are multiple print statements using the same format of 

SEPARATOR + String +SEPARATOR 

can consider making it into a function just for printing in that way

naijie2108 (commented):
Good job! I liked how you commented each method

src/main/java/Duke.java
```diff
@@ -2,58 +2,102 @@
...
+    public static void addDeadline(String line) {
+        String[] lineArray = line.substring(9).split("/by");
+        userInputs[userInputCount] = new Deadline(lineArray[0].trim(), lineArray[1].trim());
+
+        System.out.println(SEPARATOR + "\n\tadded: " + userInputs[userInputCount] + "\n" + SEPARATOR);
+        userInputCount++;
+    }
+
+    public static void addEvent(String line) {
+        String[] lineArray = line.substring(6).split("/at");
```
t-l-xin: use constant name for 6, because it is not clear why 6 is used, or what 6 means.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,296 @@
...
+
+    /**
+     * Process user input
+     * @param input the user input
+     */
+    public static void processInput(String input) {
+        input = input.trim();
+
+        try {
+            if (input.startsWith("list")) {
```
hughjazzman: Perhaps these magic strings could be declared as constants? Also, the methods could perhaps belong to another class?

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,296 @@
...
+        final String DELIMITER = "--";
+        int doneValue;
+        FileWriter beginWrite = new FileWriter("data/duke.txt");
+
+        beginWrite.write("");
+        beginWrite.close();
+
+        FileWriter fw = new FileWriter("data/duke.txt", true);
+
+        for (int i = 0; i < userInputsCount; i++) {
```
hughjazzman: The code within this loop could be improved by [Avoiding Complicated Expressions](https://nus-cs2113-ay2122s1.github.io/website/se-book-adapted/chapters/codeQuality.html#avoid-complicated-expressions). You may also consider using `isinstance`.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,296 @@
...
+        }
+
+        final int INDEX_DONE = 1;
+        int taskIndexDone = Integer.parseInt(taskInputs[INDEX_DONE]) - 1;
+
+        userInputs.get(taskIndexDone).markAsDone();
+
+        System.out.println(
+                SEPARATOR + "\n\tfinished this task... I need a break:\n\t\t" +
+                userInputs.get(taskIndexDone));
```
hughjazzman: There may be an extra newline here.
```suggestion
        System.out.println(SEPARATOR + "\n\tfinished this task... I need a break:\n\t\t" +
                userInputs.get(taskIndexDone));
```

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,296 @@
...
+        if (!line.contains("/at")) {
+            throw new DukeException("wrong input format for command event");
+        }
+
+        final int DESCRIPTION_INDEX = 0;
+        final int AT_INDEX = 1;
+        final int START_INDEX = 6;
+
+        String[] eventInputs = line.substring(START_INDEX).split("/at");
+        userInputs.add(new Event(eventInputs[DESCRIPTION_INDEX].trim(),
```
hughjazzman: Perhaps could avoid complicated expressions here.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,296 @@
...
+        final int START_INDEX = 5;
+        String[] todoInputs = line.split(" ");
+
+        if (todoInputs.length == 1) {
+            throw new DukeException("command todo description missing");
+        }
+
+        userInputs.add(new Todo(line.substring(START_INDEX)));
+
+        System.out.println(SEPARATOR + "\n\tadded: " + userInputs.get(userInputsCount) + "\n" + SEPARATOR);
```
hughjazzman: For the printing, you could consider building the strings first, before printing them, to avoid complicated expressions.

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,58 @@
...
+        return " ";
+    }
+
+    public String getBy() {
+        return " ";
+    }
+
+    public String getAt() {
+        return " ";
+    }
```
hughjazzman: These may not be needed as they return an empty string. See the comment about `isinstance`.

hughjazzman (commented):
Nice work! Do take a look at the comments. " 👍

PR #81, opened 2021-08-29
[Sean Robert den Hartog] iP

SeanRobertDH:


src/main/java/TimedTask.java
```diff
@@ -0,0 +1,20 @@
...
+        String[] inputs = userInput.split(String.format(" /%s ", preposition));
+        this.setDescription(inputs[0]);
+        this.dateTime = inputs[1];
+        this.preposition = preposition;
+    }
+
+    @Override
+    public String toString() {
+        //Append preposition and datetime
+        return String.format("%s (%s: %s)", super.toString(), preposition, dateTime);
```
irvinseet: Good use of placeholder as compared to concatenating strings. Makes code more readable.

src/main/java/TaskManager.java
```diff
@@ -0,0 +1,47 @@
...
+        Message.printWithSpacers("Nice! I've marked this task as done:\n" + currentTask);
+    }
+
+    public static void newTask(String userInput) {
+        String command = userInput.split(" ")[0];
+        //Remove command from userInput.
+        userInput = userInput.substring(command.length() + 1);
+
+        switch (Task.Types.valueOf(command.toUpperCase())) {
+        case DEADLINE:
```
irvinseet: Good indentation of switch-case statements

src/main/java/Message.java
```diff
@@ -0,0 +1,27 @@
...
+            + "|  _ \\ _   _| | _____ \n"
+            + "| | | | | | | |/ / _ \\\n"
+            + "| |_| | |_| |   <  __/\n"
+            + "|____/ \\__,_|_|\\_\\___| inc.\n";
+    private static final String INTRO = "Hello! I'm Duke\nWhat can I do for you?";
+    private static final String END = "Bye. Hope to see you again soon!";
+    private static final String SPACER = "____________________________________________________________\n";
+
+    public static void printWithSpacers(String message) {
+        if (message.charAt(message.length() - 1) != '\n') {
```
irvinseet: Can add a comment to tellothers what is happening here

src/main/java/Duke.java
```diff
@@ -1,10 +1,33 @@
+import java.util.Scanner;
+
 public class Duke {
+
+    private static final Scanner IN = new Scanner(System.in);
+
+    private static final String TERMINATE_CONSOLE = "bye";
+    private static final String LIST = "list";
+    private static final String DONE_REGEX = "done \\d+";
```
kengjit: Good choice of constants and its naming.

src/main/java/Duke.java
```diff
@@ -1,10 +1,33 @@
...
+                break;
+            } else if (userInput.equals(LIST)) {
+                TaskManager.printTasks();
+            } else if (userInput.matches(DONE_REGEX)) {
+                int id = Integer.parseInt(userInput.split(" ")[1]);
+                //id entered with index starting from '1' instead of '0'
+                TaskManager.taskDone(id - 1);
+            } else {
+                TaskManager.newTask(userInput);
+            }
```
kengjit: Follows good if-else if guidelines according to Coding Standards.

kengjit (commented):
Overall well written code. Perhaps adding some documentations on your methods would help with understanding your code better.

src/main/java/duke/Command.java
```diff
@@ -0,0 +1,85 @@
+package duke;
+
+import duke.exception.InvalidCommandException;
+import duke.task.Task;
+import duke.task.TaskManager;
+
+class Command {
+
+    private enum Commands {
```
Tejasvas123: Perhaps a better approach would be to have enums to just store the `typeOfCommands` and then have some sort of a Parser Class that handles the parsing.

src/main/java/duke/Command.java
```diff
@@ -0,0 +1,85 @@
...
+        BYE("bye", 1),
```
Tejasvas123: string literals like `"bye"`, `"list"` should be defined as named constants

src/main/java/duke/Command.java
```diff
@@ -0,0 +1,85 @@
...
+                throw new InvalidCommandException("OOPS!!! I'm sorry, but I don't know what that means :-(");
+            }
+            Commands command = Commands.valueOf(userInputSplit[0].toUpperCase());
+            if (userInputSplit.length != command.ARGS || !command.matchesRegex(userInput)) {
+                throw new InvalidCommandException(command.getUsage());
+            }
+            switch (command) {
+            case BYE:
+                return false;
+                //Fallthrough
```
Tejasvas123: Remove unnecessary comments. Also I dont see the fall through here as you are returning `false` from the BYE case

src/main/java/duke/IoManager.java
```diff
@@ -0,0 +1,70 @@
...
+public class IoManager {
+    private static final String SAVE_FOLDER = "data";
+    private static final String DIRECTORY = System.getProperty("user.dir") + '/' + SAVE_FOLDER;
+    private static final String FILE_NAME = DIRECTORY + "/duke.txt";
+
+    public static boolean init() {
+        try {
+            Path path = Paths.get(DIRECTORY);
+            Files.createDirectories(path);
+            File myObj = new File(FILE_NAME);
```
Tejasvas123: can name `myObj` to something meaningful

src/main/java/duke/IoManager.java
```diff
@@ -0,0 +1,70 @@
...
+        return loadedTasks;
+    }
+
+    public static void overwriteFile(String toWrite) {
+        try {
+            FileWriter myWriter = new FileWriter(FILE_NAME);
+            myWriter.write(toWrite);
+            myWriter.close();
+        } catch (IOException ioe) {
+            Message.printWithSpacers("Unable to Write to save file, IO exception encountered.");
```
Tejasvas123: Error messages also constitute as String literals and constants and should be added as named constants.

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,86 @@
+package duke.task;
+
+import duke.Message;
+
+public class Task {
+    private boolean isDone;
+    private String description;
+    private final Types type;
+
+    public enum Types {
```
Tejasvas123: Umm may you could explain me why do you want to store these types as `enums`. If you are inheriting everything from either `TimedTask` or `Task` then whats the need of these types?.
I would recommend this approach where lets say `Event` inherits from `TimedTask` and has all the functionalities of `Event`. With such an approach the class `Task` or `TimedTask` does not have to know that Event class exists (separation of concerns)

src/main/java/duke/task/TaskManager.java
```diff
@@ -0,0 +1,139 @@
...
+public class TaskManager {
+    public static final int MAX_TASKS = 100;
+
+    private static ArrayList<Task> tasks = new ArrayList<Task>(MAX_TASKS);
+
+    private interface Lambda {
+        void execute();
+    }
+    //TODO What do I even call this class?
+    private static class Runnable {
```
Tejasvas123: Hmm these names are not really clear. May be `runTasks` for runnable? Also this piece of code looks a bit complicated (especially with lambdas). You might want to add some comments explaining the `WHAT` and `WHY` part of the code

src/main/java/duke/task/TaskManager.java
```diff
@@ -0,0 +1,139 @@
...
+                break;
+            }
+            printTaskDone(tasks.get(tasks.size() - 1));
+            saveTasks();
+        } catch (NoDescriptionException nde) {
+            Message.printWithSpacers(nde.getMessage());
+        } catch (WrongNumberOfArgumentsException wnoae) {
+            Message.printWithSpacers(wnoae.getMessage());
+        }
+    }
```
Tejasvas123, line 149: There should be an empty line after the end of the function

src/main/java/duke/task/TimedTask.java
```diff
@@ -0,0 +1,40 @@
...
+    TimedTask(String userInput, String preposition, Types type) throws WrongNumberOfArgumentsException {
+        //TODO: Is using constructor class to parse inputs good OOP?
+        super(type);
+        String[] inputs = userInput.split(Message.WHITESPACE_REGEX + '/' + preposition + Message.WHITESPACE_REGEX);
+        if (inputs.length != 2) {
+            throw new WrongNumberOfArgumentsException(type.toString(), preposition);
+        }
+        this.setDescription(inputs[0]);
+        this.dateTime = inputs[1];
+        this.preposition = preposition;
```
Tejasvas123: Perhaps adding these to a separate function and using constructor for just initialising would be better. Also why not send directly the parsed input to the contructor? Basically why not handle the parsing in some other class and just pass the data in required format

src/main/java/duke/task/TimedTask.java
```diff
@@ -0,0 +1,40 @@
...
+    TimedTask(boolean isDone, String description, String dateTime, String preposition, Types type) {
+        //TODO: Is using constructor class to parse inputs good OOP?
+        super(isDone, description, type);
+        this.dateTime = dateTime;
+        this.preposition = preposition;
+    }
+
+    @Override
+    public String toString() {
+        //Append preposition and datetime
```
Tejasvas123: Remove redundant comments (quite clear what you are trying to do)

src/main/java/duke/task/TimedTask.java
```diff
@@ -0,0 +1,40 @@
...
+        String[] inputs = userInput.split(Message.WHITESPACE_REGEX + '/' + preposition + Message.WHITESPACE_REGEX);
+        if (inputs.length != 2) {
+            throw new WrongNumberOfArgumentsException(type.toString(), preposition);
+        }
+        this.setDescription(inputs[0]);
+        this.dateTime = inputs[1];
+        this.preposition = preposition;
+    }
+
+    TimedTask(boolean isDone, String description, String dateTime, String preposition, Types type) {
```
Tejasvas123: Yes, this seems like a good approach. Please remove the redundant comments

Tejasvas123 (commented):
Great job Sean! Added some coding quality and coding standard comments for your IP

src/main/java/duke/Command.java
SeanRobertDH: Hi Tejas, just to clarify. I've tied the use of the string literal to the enum which is a constant? If I made a named constant only to have them reference them in the enum would that not be redundant? Also, string literals can be gotten from enums using the enum's name()/tostring() method.

src/main/java/duke/task/Task.java
SeanRobertDH: I tend to use enums to store 'short strings' like commands and task names. the advantage I see is that you can still get the string in uppercase using name() method and lowercase by overwriting the toString() method. Also, from my understanding, switch and if statements become less computationally intensive with enums because you can use the typeOf() method to convert the string to an enum and compare based on the enum (which is an int as opposed to a string). Also, my Event does inherit from TimedTask. Given that 'Deadline' and 'Event' are similar except for their preposition, I used an abstract parent class TimedTask to implement the shared methods.

I also think that there's less ambiguity with comparing enums because string literals can have different case (should you call toUpper() or toLower()? you might have a different convention than your friend.) whereas since enums are always uppercase, you always do enum.valueOf(input.toUpper()).

src/main/java/duke/IoManager.java
SeanRobertDH: Would that include string literals such as "[%c][%c] %s" in the toString method of Task class? My thoughts are that it makes more sense to leave such strings in the method so it's easier to see how the format method works.

PR #83, opened 2021-08-29
[Nguyen Van Binh] iP

nvbinh15:


src/main/java/Duke.java
```diff
@@ -1,10 +1,199 @@
...
+    private static Task[] tasks = new Task[MAX_NUMBER_OF_TASKS];
+    private static int numberOfTasks = 0;
+
     public static void main(String[] args) {
-        String logo = " ____        _        \n"
-                + "|  _ \\ _   _| | _____ \n"
-                + "| | | | | | | |/ / _ \\\n"
-                + "| |_| | |_| |   <  __/\n"
-                + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
```
leowyy99, line 8: Would be good it you can add the above portion into printGreeting() as well, this will make it neater.

src/main/java/Duke.java
```diff
@@ -1,10 +1,199 @@
...
+        printGreeting();
+        Scanner in = new Scanner(System.in);
+        while (true) {
+            String userInput = getUserInput(in);
+            executeCommand(userInput);
```
leowyy99: This portion is very concise. Good job! It is a very good implementation of SLAP principle.

src/main/java/Duke.java
```diff
@@ -1,10 +1,199 @@
+import java.util.Scanner;
+
 public class Duke {
+
+    private static final String HORIZONTAL_LINE = "__________________________________________________";
```
leowyy99: Really great job on the constant naming, makes the code very easy to read.

src/main/java/Duke.java
```diff
@@ -1,10 +1,199 @@
+import java.util.Scanner;
+
 public class Duke {
```
leowyy99: Overall a very good piece of code. Easy to read and understand.

src/main/java/Duke.java
nvbinh15: Thank you for your suggestion. Good idea!

src/main/java/Duke.java
nvbinh15: Thank you.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,201 @@
...
+     * Executes the command as specified by the user input.
+     *
+     * @param userInput Raw input from user.
+     */
+    private static void executeCommand(String userInput) throws DukeException, IOException {
+        final String[] commandTypeAndParams = parser.splitCommandAndArgs(userInput);
+        final String commandType = commandTypeAndParams[0];
+        final String commandArgs = commandTypeAndParams[1];
+
+        switch (commandType) {
```
sinteary: Good use of switch case, and good abstraction of each command into separate functions! SLAP is fulfilled

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,201 @@
...
+        ui.printConfirmDelete(tasks.getTaskByIndex(taskIndex), tasks.getNumberOfTasks());
+        tasks.removeTaskByIndex(taskIndex);
+    }
+
+    /**
+     * Executes listing tasks.
+     */
+    private static void executeList() {
+        System.out.println(HORIZONTAL_LINE);
+        if (tasks.getNumberOfTasks() > 0) {
```
sinteary: You can make the [happy path](https://nus-cs2113-ay2122s1.github.io/website/se-book-adapted/chapters/codeQuality.html#make-the-happy-path-prominent) more prominent here. Eg. check if number of tasks <= 0, if yes, print the MESSAGE_NO_TASK then `return`. Then the rest of your code to handle the subsequent logic doesn't need to be nested inside `if else` blocks. Similar thing can be done for `executeAddEvent` and some other similar functions here

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,201 @@
...
+        System.exit(0);
+    }
+
+    /**
+     * Executes marking task as done.
+     *
+     * @param commandArgs Processed input arguments from user.
+     */
+    private static void executeMarkDone(String commandArgs) {
+        int taskIndex = Integer.parseInt(commandArgs.strip()) - 1;
```
sinteary: Good and clear naming of variables! Likewise for most of the other variables I see in your functions

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,201 @@
...
+     *
+     * @param commandArgs Processed input arguments from user.
+     */
+    private static void executeAddEvent(String commandArgs) throws IllegalEventException {
+        int indexOfTimePrefix = commandArgs.indexOf(PREFIX_TIME_EVENT);
+        if (indexOfTimePrefix != -1) {
+            String description = commandArgs.substring(0, indexOfTimePrefix).trim();
+            String time = commandArgs.substring(indexOfTimePrefix + 3).trim();
+            tasks.addTask(new Event(description, time));
+            ui.printConfirmAdd(tasks.getTaskByIndex(tasks.getNumberOfTasks() - 1), tasks.getNumberOfTasks());
```
sinteary: This line is a bit long, and you also repeated the function call for `tasks.getNumberOfTasks()`. You can extract this into a variable eg. `numberOfTasks` and reuse it.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,90 @@
...
+        } else {
+            prefix = PREFIX_TODO;
+            data = prefix + VERTICAL_BAR + statusIcon + VERTICAL_BAR + description + LINE_SEPARATOR;
+        }
+        return data;
+    }
+
+    public static Task retrieveStoredData(String data) {
+        String[] tokens = data.split(VERTICAL_BAR_REGEX);
+        String prefix = tokens[0];
```
sinteary: Extracting these and naming them as variables makes the code more readable and easy to understand, good job!

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,90 @@
...
+        String statusIcon = tokens[1];
+        String description = tokens[2];
+        boolean isDone = (statusIcon.equals(STATUS_ICON_DONE)) ? true : false;
+
+        if (prefix.equals(PREFIX_EVENT)) {
```
sinteary: Possible to use a switch case with a default branch here

src/main/java/duke/constants/Messages.java
```diff
@@ -0,0 +1,34 @@
...
+
+public class Messages {
+
+    public final static String LOGO = "\t" + " ____        _\n"
+            + "\t" + "|  _ \\ _   _| | _____\n"
+            + "\t" + "| | | | | | | |/ / _ \\\n"
+            + "\t" + "| |_| | |_| |   <  __/\n"
+            + "\t" + "|____/ \\__,_|_|\\_\\___|\n\n";
+
+    public static final String OPEN_SQUARE_BRACKET = "[";
```
sinteary, line 14: Good effort extracting all the constants!

PR #84, opened 2021-08-29
[Ma Zijian] iP

MAZJ124:


src/main/java/Duke.java
```diff
@@ -1,10 +1,110 @@
...
+
+    private static final String LOGO = " ____        _        \n"
+                                    + "|  _ \\ _   _| | _____ \n"
+                                    + "| | | | | | | |/ / _ \\\n"
+                                    + "| |_| | |_| |   <  __/\n"
+                                    + "|____/ \\__,_|_|\\_\\___|\n";
+    private static final String LINE_SEPARATOR = "------------------------------------";
+    private static final String HELLO_MESSAGE = "Hello! I'm Duke\nWhat can I do for you?";
+    private static final String BYE_MESSAGE = "Bye. Hope to see you again soon!";
+    private static final String ERROR_MESSAGE = "You need to specify the task type!";
```
KVignesh122: Good use of constants.

src/main/java/Event.java
```diff
@@ -0,0 +1,15 @@
...
+        this.taskType = 'E';
+        this.at = at;
+    }
+
+    @Override
+    public String getTaskInfo() {
+        String date = at.split(" ", 2)[1];
+        return super.getTaskInfo() + "(at: " + date + ")";
+    }
+}
```
KVignesh122: Clean and neat code. No coding violations found. :)

src/main/java/Task.java
```diff
@@ -0,0 +1,31 @@
...
+        isDone = true;
+    }
+
+    public String getTaskInfo() {
+        return getTaskType() + getStatus() + description.split(" ", 2)[1];
+    }
+
+    public void printTask(int index){
+        System.out.println(index + "." + getTaskInfo());
+    }
```
KVignesh122: All method names correctly follow camelCase.

src/main/java/Duke.java
```diff
@@ -1,10 +1,110 @@
...
+                break;
+            case "event":
+                addEvent(input);
+                break;
+            case "done":
+                finishTask(input);
+                break;
+            default:
+                System.out.println(ERROR_MESSAGE);
+                break;
```
KVignesh122: switch-case statement indentation is correct.

KVignesh122 (commented):
Overall good job. Good coding logic.

src/main/java/Duke.java
```diff
@@ -1,10 +1,110 @@
...
     public static void main(String[] args) {
-        String logo = " ____        _        \n"
-                + "|  _ \\ _   _| | _____ \n"
-                + "| | | | | | | |/ / _ \\\n"
-                + "| |_| | |_| |   <  __/\n"
-                + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
+        printHelloMessage();
+        handleCommand();
+        printByeMessage();
```
serenakemono: I like how you give methods clear names which make your code very clear.

src/main/java/Duke.java
```diff
@@ -1,10 +1,110 @@
...
+
+    private static void handleCommand() {
+        String input;
+        Scanner in = new Scanner(System.in);
+        input = in.nextLine();
+        while (!input.equals("bye")) {
+            String keyword = input.split(" ")[0].toLowerCase();
+            switch (keyword) {
+            case "bye":
+                continue;
```
serenakemono: I'm not too sure if this case is relevant since it is already accounted for in line 29.

src/main/java/Event.java
```diff
@@ -0,0 +1,15 @@
+public class Event extends Task {
+    protected String at;
+
+    public Event(String description, String at) {
+        super(description);
+        this.taskType = 'E';
+        this.at = at;
+    }
+
+    @Override
```
serenakemono: I like the clear indication of overridden method.

src/main/java/Duke.java
```diff
@@ -1,10 +1,110 @@
...
+    }
+
+    private static void printHelloMessage() {
+        System.out.println("Hello from\n" + LOGO);
+        System.out.println(LINE_SEPARATOR);
+        System.out.println(HELLO_MESSAGE);
+        System.out.println(LINE_SEPARATOR);
+    }
+
+    private static void printByeMessage() {
```
serenakemono: I like the namings of these two methods. They sound similar suggesting the similarity of their nature.

serenakemono (commented):
I really like how succinct and neat your codes are. Good job!

src/main/java/Duke.java
MAZJ124: Thanks for pointing out! I believe this statement is still required since it will break me out of the current loop and bring me to the next loop (which then checks whether my input is "bye" and exits the whole function).

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,261 @@
...
+            dir.mkdir();
+            File newFile = new File("data/duke.txt");
+            newFile.createNewFile();
+            return newFile;
+        }
+        return file;
+    }
+
+    private static void addTasksIntoList() throws IOException {
+        File f = loadFile();
```
Mick609: Please consider use name like "file" for the File object instead of "f". Similar problems can be found from other variables as well.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,261 @@
...
+        System.out.println(LINE_SEPARATOR);
+        System.out.println(HELLO_MESSAGE);
+        System.out.println(LINE_SEPARATOR);
+    }
+
+    private static void printByeMessage() {
+        System.out.println(BYE_MESSAGE);
+        System.out.println(LINE_SEPARATOR);
+    }
+}
```
Mick609, line 77: You have packed most of the methods inside this one class. Please consider putting some methods in other classes. For example, all the task-related methods can be under a new class called TaskManager.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,261 @@
...
+        boolean noInput = inputSplit.length == 1;
+        if (noInput) {
+            throw new DukeException("You have to specify the task!");
+        }
+        boolean noSeparator = !input.contains("/");
+        if (noSeparator) {
+            throw new DukeException("Your task is not in the right format");
+        }
+        String[] inputWords = input.split("/");
+        boolean noTask = inputWords[0].trim().equals("deadline");
```
Mick609: Please consider changing the boolean variable's name to isNoTask.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,261 @@
...
+    private static void saveTasks() {
+        try {
+            writeTasksToFile();
+        } catch (IOException e) {
+            System.out.println("Something went wrong: " + e.getMessage());
+        }
+    }
+
+    private static void writeTasksToFile() throws IOException {
+        FileWriter fw = new FileWriter(FILE_PATH);
```
Mick609: Please consider using names like "fileWriter" for the FileWriter object instead of "fw".

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,261 @@
...
+
+    private static void handleCommand() {
+        String input;
+        Scanner in = new Scanner(System.in);
+        input = in.nextLine();
+        while (!input.equals("bye")) {
+            try {
+                String keyword = input.split(" ")[0].toLowerCase();
+                switch (keyword) {
+                case "bye":
```
Mick609, line 33: When the user types the "bye" command with some extensions (so the command is not matching the "bye" String in the condition of the while loop), isn't the user still want to close the program? Or maybe we should prompt something to the user instead of ignoring it?

Mick609 (commented):
The code is clean and easy to understand with minor naming convention issues.

PR #93, opened 2021-08-29
[Duc Anh Tran Nhan] iP

datn02:


src/main/java/Parser.java
```diff
@@ -0,0 +1,161 @@
+public class Parser {
+    static private final String COMMAND_LIST = "list";
```
jiangweichen835: I think the declaration of the constant names makes your code look neat.

src/main/java/Ui.java
```diff
@@ -0,0 +1,32 @@
+import java.util.Scanner;
+
+public class Ui {
+    private static final String SEPARATING_LINE = "\t-------------------------------------";
+    private static final String GREET_MESSAGE = "Yo what's up I'm Jesse. Need help?";
+    private static final String BYE_MESSAGE = "So long!";
+
+    Scanner scanner = new Scanner(System.in);
```
jiangweichen835: perhaps change the variable name scanner to another name?

src/main/java/Ui.java
```diff
@@ -0,0 +1,32 @@
...
+
+    public void greet() {
+        print(GREET_MESSAGE);
+    }
+
+    public void bye() {
+        print(BYE_MESSAGE);
+    }
+
+    public Ui() {
```
jiangweichen835: Was this a mistake? Should there be any code in this method?

src/main/java/Duke.java
```diff
@@ -1,10 +1,25 @@
...
+
 public class Duke {
     public static void main(String[] args) {
-        String logo = " ____        _        \n"
-                + "|  _ \\ _   _| | _____ \n"
-                + "| | | | | | | |/ / _ \\\n"
-                + "| |_| | |_| |   <  __/\n"
-                + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
+        Ui ui = new Ui();
```
jiangweichen835: Perhaps change the variable name to another name instead of ui ui?

jiangweichen835 (commented):
I think overall good coding standard! I'm sorry there wasn't much constructive reviews from me.

src/main/java/Parser.java
```diff
@@ -0,0 +1,161 @@
...
+
+        userTaskIndex++;
+        return msg;
+    }
+
+    private static String parseDoneCommand(String command) {
+        String msg;
+        String detail = command.substring(COMMAND_DONE.length()).trim();
+
+        int taskDoneNumber = Integer.parseInt(detail);
```
Kair0s3: I feel that the naming of the variable is a little confusing here as it can be interpreted differently eg. is it "a counter for no. of Task that has been done" or "the index of the task in the list that you intend to mark as done"

src/main/java/ToDos.java
```diff
@@ -0,0 +1,9 @@
+public class ToDos extends Task {
```
Kair0s3: Would it be less confusing if the class name is singular instead of plural?

src/main/java/Parser.java
```diff
@@ -0,0 +1,161 @@
...
+
+    public boolean isExit(String command) {
+        return (command.equals(COMMAND_EXIT));
+    }
+
+    private static String parseListCommand() {
+        String msg = "";
+
+        if (userTaskIndex == 0) {
+            msg = "Nothing.";
```
Kair0s3: Maybe the msg could be a constant instead?

src/main/java/Duke.java
```diff
@@ -1,10 +1,25 @@
+import java.util.Scanner;
```
Kair0s3: In general, I quite like that the "Main" (in this case, Duke) code seems very clean as most of the processing/parsing functions are abstracted on a high level.

src/main/java/Parser.java
```diff
@@ -0,0 +1,161 @@
...
+    private static String parseDoneCommand(String command) {
+        String msg;
+        String detail = command.substring(COMMAND_DONE.length()).trim();
+
+        int taskDoneNumber = Integer.parseInt(detail);
+
+        if (taskDoneNumber <= 0 || taskDoneNumber > userTaskIndex) {
+            msg = "Task number not exist!";
+            return msg;
+        }
```
Kair0s3: Good use of happy path to deal with unusual conditions of the task index to be marked

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,36 @@
+package duke;
+
+public class Duke {
+    public static void main(String[] args) {
+        Ui ui = new Ui();
+        Parser parser = new Parser();
+        String command;
```
sinteary: Why not initialize this in the same line (line 13)? Likewise for `msg` in line 27

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,235 @@
...
+import java.util.ArrayList;
+
+import duke.task.Deadline;
+import duke.task.Event;
+import duke.task.Task;
+import duke.task.ToDos;
+
+import java.io.IOException;
+
+public class Parser {
```
sinteary: It seems like `Parser` class doesn't only parse commands, but executes them as well. In that case, it might be better to change this class name to something less misleading

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,235 @@
...
+    static private final String EMPTY_STRING = "";
+
+    static private final int EVENT_DEADLINE_ARGUMENT_COUNT = 2;
+
+
+    private static ArrayList<Task> userTasks;
+    private static TaskStorage storage;
+
+    public Parser() {
+        //userTasks = new ArrayList<>();
```
sinteary: You should remove commented code

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,235 @@
...
+        } catch (IOException e) {
+            System.out.println("Something went wrong in file creation :(");
+        }
+    }
+
+
+    public static String parse(String command) throws DukeException {
+        String[] words = command.split(SPACE_SEPARATOR);
+
+        if (words[0].equals(COMMAND_LIST)) {
```
sinteary: Might be better (and also more readable!) to use `switch case` statements here, instead of checking if else for each of them.

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,235 @@
...
+        } else if (words[0].equals(COMMAND_DELETE)) {
+            return parseDeleteCommand(command);
+        }
+        else {
+            throw new DukeException(INVALID_COMMAND);
+        }
+    }
+
+    private static String parseDeleteCommand(String command) throws DukeException{
+        String msg;
```
sinteary: Maybe `outputMessage` is clearer, to show that this function returns an output message

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,235 @@
...
+        if (detail.length() <= 0 || contentAndDate.length <= 0) {
+            throw new DukeException("Invalid format. Enter by this format:\n"
+                    + "\t\t\"deadline [description] /by [deadline]\"");
+        }
+
+        for (int i = 0; i < contentAndDate.length; i++) {
+            if (contentAndDate.length != EVENT_DEADLINE_ARGUMENT_COUNT
+                    || contentAndDate[i].equals(EMPTY_STRING)) {
+                throw new DukeException("Invalid format. Enter by this format:\n"
+                        + "\t\t\"deadline [description] /by [deadline]\"");
```
sinteary: This seems to be an exception with a message similar to line 142: maybe it's possible to extract this string out as a constant somewhere to avoid repeating yourself. Likewise for some of your other exception messages

src/main/java/duke/Parser.java
```diff
@@ -0,0 +1,235 @@
...
+    }
+
+    private static String parseDoneCommand(String command) throws DukeException {
+        String msg;
+        String detail = command.substring(COMMAND_DONE.length()).trim();
+
+        if (detail.length() <= 0) {
+            throw new DukeException("Bro don't you say you've done all that.\n" +
+                    "\tPick a number with your done task!");
+        } else {
```
sinteary: You don't really need the `else` here: since you will already have thrown the exception if the condition is not met. You can then avoid nesting everything else in the `else` block
Removing the `else` will make the [happy path](https://nus-cs2113-ay2122s1.github.io/website/se-book-adapted/chapters/codeQuality.html#make-the-happy-path-prominent) clearer

src/main/java/duke/TaskStorage.java
```diff
@@ -0,0 +1,133 @@
...
+
+        for (Task t : tasks) {
+            data += taskToString(t) + System.lineSeparator();
+        }
+
+        fw.write(data);
+        fw.close();
+    }
+
+    private String taskToString(Task t) {
```
sinteary: This function can be better named eg. `getStorageFormatString`.
Also, you can consider using inheritance and overriding methods to make this part slightly easier for you (similar to the method `toString`)

src/main/java/duke/TaskStorage.java
```diff
@@ -0,0 +1,133 @@
...
+            Deadline deadlineTask = (Deadline) t;
+            return String.format("deadline|%s|%s|%b", deadlineTask.getTaskDescription(),
+                    deadlineTask.getDate(), deadlineTask.isDone());
+        } else {
+            return null;
+        }
+    }
+
+    private Task readTask(String string) throws DukeException {
+        String[] detail = string.split(SPACE_SEPARATOR);
```
sinteary: `details` would better reflect the plural parts. Also the input parameter could be better named as `taskString` or `taskDetailsString` to make it clearer

PR #95, opened 2021-08-29
[Chew Yi Jie] iP

ChewyChair:


src/main/java/Shika.java
```diff
@@ -0,0 +1,191 @@
+import java.util.Scanner;
+
+public class Shika {
+
+    public static String line = "____________________________________________________________________________\n";
```
theeugenechong: This variable could be declared as `static final` as it is a constant String.

src/main/java/Shika.java
```diff
@@ -0,0 +1,191 @@
...
+
+    /**
+     * Main function that calls other functions to run Shika.
+     * @param args Command line arguments.
+     */
+    public static void main(String[] args) {
+        greetUser();
+        runShika();
+    }
```
theeugenechong: I like how clean and simple your `main()` method is! 👍Good job!

src/main/java/Shika.java
```diff
@@ -0,0 +1,191 @@
...
+     * @param taskList Array containing all recorded tasks.
+     * @param text String containing user input.
+     */
+    public static void addDeadline(Task[] taskList, String text) {
+        if (!text.contains("/by")) {
+            System.out.print(line + "Please follow the format [NAME] /by [DEADLINE]. Thank you!\n" + line);
+            return;
+        }
+        String str = text.substring(text.indexOf("deadline") + 8, text.indexOf("/")).trim();
+        String by = text.substring(text.indexOf("/by") + 3).trim();
```
theeugenechong: Perhaps you could introduce constants (for e.g. DEADLINE_LENGTH) to define what the "magic numbers" 8 and 3 are. Noticed this in the other methods `addEvent`, `doTask` as well.

theeugenechong (commented):
Overall, you have very clean code which is broken down into different levels of abstraction and then put back together in a very organized manner! 💯

src/main/java/shikabot/ui/TextUi.java
```diff
@@ -0,0 +1,116 @@
...
+    /**
+     * Function to print all tasks in tasks.
+     */
+    public void printTasks(TaskList taskList) {
+        System.out.println(line + "> Here is your list of tasks: ") ;
+        for (int i = 0; i < taskList.size(); i++) {
+            System.out.println("\t" + (i + 1) + ". " + taskList.get(i).toString());
+        }
+        printTaskCount(taskList.size());
+    }
```
EthanWong2212: Perhaps it would be better to make all methods static methods so that you will not need to declare a TextUI object to call these methods

src/main/java/shikabot/command/FailedCommand.java
```diff
@@ -0,0 +1,46 @@
...
+            ui.printInvalidTaskMessage();
+            break;
+        case (EMPTY_FIELD):
+            ui.printEmptyFieldMessage();
+            break;
+        default:
+            break;
+        }
+    }
+}
```
EthanWong2212: nice way to handle errors

src/main/java/shikabot/parser/Parser.java
```diff
@@ -0,0 +1,142 @@
...
+    public static final int NEGATIVE_INDEX_ERROR = 4;
+    public static final int INVALID_TASK = 5;
+    public static final int EMPTY_FIELD = 6;
+
+    private static final int DONE_LENGTH = 4;
+    private static final int TODO_LENGTH = 4;
+    private static final int DEADLINE_LENGTH = 8;
+    private static final int EVENT_LENGTH = 5;
+    private static final int DELETE_LENGTH = 6;
+    private static final int DIVIDER_LENGTH = 3;
```
EthanWong2212: good use of constants to avoid use of magic numbers

src/main/java/shikabot/parser/Parser.java
```diff
@@ -0,0 +1,142 @@
...
+            command = new ListCommand();
+        } else if (text.startsWith("done")) {
+            command = parseDoneCommand(text);
+        } else if (text.startsWith("delete")) {
+            command = parseDeleteCommand(text);
+        } else if (isAddCommand(text)) {
+            command = parseAddCommand(text);
+        } else {
+            command = new InvalidCommand();
+        }
```
EthanWong2212: perhaps it would look neater if you use switch case instead of ifelse. This is fine too

src/main/java/shikabot/parser/Parser.java
```diff
@@ -0,0 +1,142 @@
...
+            } catch (EmptyFieldException e) {
+                return new FailedCommand(EMPTY_FIELD);
+            }
+        }
+        return command;
+    }
+
+    private Command parseAddTodoCommand(String text) throws EmptyFieldException {
+        String name = text.substring(text.indexOf("todo") + TODO_LENGTH).trim();
+        if (name.equals("")) throw new EmptyFieldException();
```
EthanWong2212: Perhaps if-statements and their outcomes should be done on separate lines

src/main/java/shikabot/parser/Parser.java
```diff
@@ -0,0 +1,142 @@
...
+        return new AddCommand('T', name, "");
+    }
+
+    private Command parseAddDeadlineCommand(String text) throws EmptyFieldException, Task.InvalidTaskException {
+        if (!text.contains("/by")) {
+            throw new Task.InvalidTaskException();
+        }
+        String name = text.substring(text.indexOf("deadline") + DEADLINE_LENGTH, text.indexOf("/")).trim();
+        String by = text.substring(text.indexOf("/by") + DIVIDER_LENGTH).trim();
```
EthanWong2212: Perhaps strings in here can be in enums or constants variables to avoid the use of magic strings

src/main/java/shikabot/task/TaskList.java
```diff
@@ -0,0 +1,54 @@
+package shikabot.task;
+
+import java.util.ArrayList;
+
+public class TaskList {
+
+    public ArrayList<Task> taskList;
```
EthanWong2212: Perhaps this should be protected or private. It would be better to use getters and setters to gain access to the variable

src/main/java/shikabot/task/TaskList.java
```diff
@@ -0,0 +1,54 @@
...
+        return taskList.size();
+    }
+
+    public Task get(int index) {
+        return taskList.get(index);
+    }
+
+    public void remove(int index) {
+        taskList.remove(index);
+    }
```
EthanWong2212: Maybe it would be better to give clearer method names. (e.g. removeTask)

EthanWong2212 (commented):
Nice work overall. Can see good use of SLAP. Some minor improvements can be done. Keep up the great work

PR #98, opened 2021-08-29
[Marcus Ong Yih] iP

marcusbory:


src/main/java/TaskManager.java
```diff
@@ -0,0 +1,134 @@
+public class TaskManager {
+    private Task[] taskList;
```
ng-andre: Plural form should be used on names representing a collection of objects

src/main/java/TaskManager.java
```diff
@@ -0,0 +1,134 @@
...
+    public void shutdownMessage() {
+        System.out.println(
+                line +
+                " Very well sir, I shall leave you to your own devices.\n" +
+                line
+        );
+    }
+
+    public void processInput(String userInput) {
+        String[] destructuredInput = userInput.split(" ");
```
ng-andre: Plural form should be used on names representing a collection of objects
```suggestion
        String[] destructuredInputs = userInput.split(" ");
```

src/main/java/TaskManager.java
```diff
@@ -0,0 +1,134 @@
+public class TaskManager {
+    private Task[] taskList;
+    private int listIndex;
+
+    public TaskManager() {
+        taskList = new Task[100];
```
ng-andre: a constant can be used instead

ng-andre (commented):
Code is very readable with very minor coding violations, otherwise LGTM

src/main/java/TaskManager.java
3m0W33D: You could consider changing this to `taskLists` instead of `taskList` as this is a collective object

src/main/java/TaskManager.java
```diff
@@ -0,0 +1,134 @@
...
+
+    private void completeTask(String userInput) {
+        String[] destructuredInput = userInput.split(" ");
+        int index = Integer.parseInt(destructuredInput[1]) - 1;
+        taskList[index].setTaskDone();
+        completeTaskMessage(index+1, taskList[index].toString());
+    }
+
+    private void addTodo(String userInput) {
+        String[] destructuredInput = userInput.split(" ", 2);
```
3m0W33D: You could consider changing this to `destructuredInputs` to keep consistency with the other portions of the code when you used `destructuredArguments`

src/main/java/TaskManager.java
```diff
@@ -0,0 +1,134 @@
...
+    private Task[] taskList;
+    private int listIndex;
+
+    public TaskManager() {
+        taskList = new Task[100];
+        listIndex = 0;
+        initAlfred();
+    }
+
+    private String line = "____________________________________________________________\n";
```
3m0W33D: Consider moving the moving this variable to the top along with other variables.
In addition, since this is a constant you might want to consider changing this to `public static final String LINE = "...."`

src/main/java/TaskManager.java
```diff
@@ -0,0 +1,134 @@
...
+
+    private void listTasks() {
+        System.out.print(line);
+        if (listIndex == 0) {
+            System.out.println(" Your schedule is clear, Master Wayne.");
+        } else {
+            System.out.println(" Your tasks, sir:");
+            for (int i = 0; i < listIndex; i++) {
+                System.out.println(" " + (i + 1) + "." + taskList[i].toString());
+            }
```
3m0W33D: Good work on following the bracketing conventions!
https://se-education.org/guides/conventions/java/basic.html#conditionals

3m0W33D (commented):
Good job with the overall code a few minor convention errors but very well done!

src/main/java/TaskManager.java
```diff
@@ -76,17 +84,7 @@ private void completeTask(String userInput) {
...
-        System.out.println("    " + destructuredInput[1] + "." + taskList[index].toString());
-        System.out.println(line);
-    }
-
-    private void addTaskMessage(Task t) {
-        System.out.print(line);
-        System.out.println(" I shall put this in your schedule, Master Wayne: \n    " + t.toString());
-        System.out.println(" Sir, the number of Tasks you have scheduled currently amounts to " + listIndex + ".");
-        System.out.println(line);
+        completeTaskMessage(index+1, taskList[index].toString());
```
Herrekt: Nice attempt in abstracting these lines to another method. However, method descriptions could be added to allow for better understanding of what occurs here

Herrekt: I understand the use of index + 1, but would it be better to adjust the underlying array, so that it could just be index instead? Also, not supposed to mention coding standards but index + 1 would be preferred over index+1

src/main/java/TaskManager.java
```diff
@@ -0,0 +1,134 @@
+public class TaskManager {
```
Herrekt: As mentioned previously, it could be better to name it taskLists, but I was wondering if moving this list into the tasks itself as a static variable would have the same effect

src/main/java/Task.java
```diff
@@ -0,0 +1,26 @@
+public class Task {
+    protected String description;
+    protected boolean isDone;
+
+    public Task(String description) {
+        this.description = description;
+        this.isDone = false;
```
Herrekt: I noticed that there is no getter for isDone, and seeing that it is protected, the variable is still accessible by child/parent classes. However for the strongest form of encapsulation, private would do the trick.

Herrekt (commented):
Overall good work in the coding quality. I noticed some issues/disagreements and added comments for consideration.

src/main/java/TaskManager.java
marcusbory: Noted, since this is an array of `Task`s, I will be changing it to `tasks`.

src/main/java/TaskManager.java
marcusbory: Noted, thanks.

src/main/java/TaskManager.java
marcusbory: Noted, thanks.

src/main/java/TaskManager.java
marcusbory: Noted, I understand the initial confusion in my naming consistency. My rationale was that `destructuredInputs` will be used for splitting the user's input to decipher the **command**, while `destructuredArguments` will be used for splitting the **command arguments** (ie description of the Deadline, and the time given for the deadline). I have made some minor updates to address this. Thanks.

src/main/java/TaskManager.java
marcusbory: Noted, thanks.

src/main/java/TaskManager.java
marcusbory: My rationale for creating a `TaskManager` class was for the purpose of being able to manage my list of `Task`s intuitively in a single interface. Personally, I feel that moving the list into the `Task` class as suggested would not be very intuitive.

src/main/java/alfred/manager/DataManager.java
```diff
@@ -0,0 +1,92 @@
...
+        try {
+            if (!taskFile.exists()) {
+                taskFile.getParentFile().mkdirs();
+            }
+            if (taskFile.createNewFile()) {
+                MessageManager.createNewFileMessage();
+                return false;
+            }
+        } catch (IOException e) {
+            System.out.println("IO Exception at hasExistingFile()");
```
Mick609: Maybe printing the exact error message is a better implementation, i.e., printing the content of variable e would reveal more information for the user when an error is caught.

src/main/java/alfred/manager/DataManager.java
```diff
@@ -0,0 +1,92 @@
...
+        return true;
+    }
+
+    public static void populateTasks(ArrayList<Task> tasks) {
+        if (!hasExistingFile()) {
+            return;
+        }
+        int index = 0;
+        try {
+            Scanner s = new Scanner(taskFile);
```
Mick609: Maybe you could use a different name for the Scanner object rather than s?

src/main/java/alfred/manager/DataManager.java
```diff
@@ -0,0 +1,92 @@
...
+            return;
+        }
+        if (destructuredInputs[1].equals("true")) {
+            tasks.get(index).setTaskDone();
+        }
+    }
+
+    public static void saveAllTasks(ArrayList<Task> tasks) {
+        try {
+            FileWriter fw = new FileWriter(FILE_PATH);
```
Mick609: Again, it would be better if the name of the variable is a bit more descriptive. For example in this case, maybe fileWriter instead of fw.

src/main/java/alfred/manager/TaskManager.java
```diff
@@ -0,0 +1,145 @@
...
+        try {
+            switch (destructuredInputs[0]) {
+            case "list":
+                listTasks();
+                break;
+            case "done":
+                completeTask(userInput);
+                break;
+            case "todo":
+                // Fallthrough
```
Mick609: The cases for "event" and "todo" are falling through, the method does not do anything when these two commands show up. However, the methods for addTodo and addEvent are implemented and used in the addTask method. Is this fallthrough a mistake or the methods for Todo and Event are not ready yet?

src/main/java/alfred/task/Todo.java
```diff
@@ -0,0 +1,20 @@
...
+    public Todo(String description) {
+        super(description);
+    }
+
+    public String getType() {
+        return "T";
+    }
+
+    public String getDate() {
+        return "";
```
Mick609: If the Todo class in this context has no "date" as in the Deadline or Event, why would this class have a method getDate that returns ""? Maybe the getDate method in the parent class (Task) does not have to be an abstract method?

Mick609 (commented):
Very clean and easy to read and understand with only minor coding standards or naming convention issues.

src/main/java/alfred/manager/DataManager.java
marcusbory: Hi Mick, noted on this. Thanks!

src/main/java/alfred/manager/TaskManager.java
marcusbory: Hi Mick, I was concerned with the length of my `processInput()` function. Initially, I had implemented `addTodo()`, `addEvent()`, and `addDeadline()` within `processInput()`. However, due to exceptions handling, the length of my `processInput()` function exceeded 30 lines, and I decided to extract these three functions and handle them separately in `addTask()`. For this case, I specified a fallthrough so that the case of Todos, Events, and Deadlines, will be only handled by `addTask()`.

src/main/java/alfred/task/Todo.java
marcusbory: Hi Mick, the rationale behind having an abstract `getDate()` method was so that I am able to call `Task.getDate()` when encoding tasks to be saved in my `.txt` file, and do not have to write separate methods for encoding different types of tasks like `Todo`s, `Event`s, and `Deadline`s. May I check if this is acceptable? If not, I would have to create more methods to handle tasks with dates separately.

src/main/java/alfred/manager/TaskManager.java
Mick609: I understand. Your rationale is valid to me.

src/main/java/alfred/task/Todo.java
Mick609: I understand. I think your response makes sense to me. Thank you for the prompt response.

PR #104, opened 2021-08-29
[Muhammad Ashraf] iP

ashrafjfr:


src/main/main/java/Duke.java
```diff
@@ -32,7 +32,7 @@ public static void main(String[] args) {
                 System.out.println("____________________________________________________________\n"
                         + "Here are the tasks in your list:");
                 for(int i = 0; i < listCount; i++){
```
Cesare4869: Coding, standard. I think there should be space between "for" and braket, as well as the bracket and curly bracket

src/main/main/java/Duke.java
```diff
@@ -46,16 +46,38 @@ public static void main(String[] args) {
                 list[taskNum-1].markAsDone();
                 System.out.println("____________________________________________________________\n"
                         + "Nice! I've marked this task as done:");
-                System.out.println(" [" + list[taskNum-1].getStatusIcon() + "] " + list[taskNum-1].description);
+                System.out.println(" " + list[taskNum-1].toString());
                 System.out.println("____________________________________________________________\n");
                 continue;
```
Cesare4869: I am not sure that what do you mean skip the for loop for using the "continue" here.

src/main/main/java/Duke.java
```diff
@@ -46,16 +46,38 @@ public static void main(String[] args) {
...
+                String description = line.substring(5);
+                Todo t = new Todo(description);
+                list[listCount] = t;
+                listCount++;
+            } else if (line.toLowerCase().contains("deadline")) {
+                int deadlineDividerPosition = line.indexOf("/by");
+                String description = line.substring(9,deadlineDividerPosition);
+                String by = line.substring(deadlineDividerPosition+4);
+                Deadline t = new Deadline(description,by);
+                list[listCount] = t;
```
Cesare4869: just a comment, i prefer to use list[listCount++]

src/main/main/java/Duke.java
```diff
@@ -46,16 +46,38 @@ public static void main(String[] args) {
...
+            }   else if (line.toLowerCase().contains("event")) {
+                int eventDividerPosition = line.indexOf("/at");
+                String description = line.substring(6,eventDividerPosition);
+                String at = line.substring(eventDividerPosition+4);
+                Event t = new Event(description,at);
+                list[listCount] = t;
+                listCount++;
+            }
+
             System.out.println("____________________________________________________________\n"
```
Cesare4869: i think you can make a constant to take place of println("                          "\n), and this skips two lines?

src/main/main/java/Todo.java
```diff
@@ -0,0 +1,14 @@
...
+    protected String by;
+
+    public Todo(String description) {
+        super(description);
+    }
+
+    @Override
+    public String toString() {
+        return "[T]" + super.toString();
+    }
```
Cesare4869: all the rest is fine

Cesare4869 (commented):
good job

src/main/java/jarvis/Jarvis.java
```diff
@@ -0,0 +1,179 @@
...
+    public static final String COMMAND_BYE = "bye";
+    public static final String COMMAND_LIST = "list";
+    public static final String COMMAND_DONE = "done";
+    public static final String COMMAND_TODO = "todo";
+    public static final String COMMAND_DEADLINE = "deadline";
+    public static final String COMMAND_EVENT = "event";
+
+    public static void main(String[] args) {
+        greetMessage();
+        Scanner in = new Scanner(System.in);
```
Mick609: Maybe consider changing the name for the Scanner object to scanner or inScanner instead of "in".

src/main/java/jarvis/Jarvis.java
```diff
@@ -0,0 +1,179 @@
...
+        String userLine;
+        ArrayList<Task> taskList = new ArrayList<>();
+        Boolean isOn = true;
+        while (isOn) {
+            userLine = in.nextLine().trim();
+            if (userLine.toLowerCase().contains(COMMAND_BYE)) {
```
Mick609: Using the contains method in this situation to detect "Bye" is a bit dangerous. What if the user's new Todo is "Say Goodbye to someone"?

src/main/java/jarvis/Jarvis.java
```diff
@@ -0,0 +1,179 @@
...
+                + "You currently have " + taskList.size() + " tasks in your list Sir.\n"
+                + LINE);
+    }
+
+    private static String extractTask(String userLine) {
+        String[] splitString = userLine.toLowerCase().split(" ",2);
+        return splitString[0];
+    }
+
+    private static void greetMessage() {
```
Mick609: Please consider changing the name of this method to "printGreetMessage" as an action. Similar problems can be found in the following methods.

src/main/java/jarvis/Jarvis.java
```diff
@@ -0,0 +1,179 @@
...
+    private static void printAddedTask(Task task, ArrayList<Task> taskList) {
+        System.out.println(LINE_W_NL
+                + "Noted Sir. I've added this task:\n"
+                + " " + task.toString() + "\n"
+                + "You currently have " + taskList.size() + " tasks in your list Sir.\n"
+                + LINE);
+    }
+
+    private static String extractTask(String userLine) {
+        String[] splitString = userLine.toLowerCase().split(" ",2);
```
Mick609: Just a suggestion: adding an error exception catch statement here will help the software works more stable. As the userLine here contains the user input (which can be anything). the result of the splitting method may as well be empty. In that case, the error will crash the software.

src/main/java/jarvis/Jarvis.java
```diff
@@ -0,0 +1,179 @@
...
+                deadlineTask(userLine, taskList);
+                break;
+            case COMMAND_EVENT:
+                eventTask(userLine, taskList);
+                break;
+            case COMMAND_BYE:
+                byeMessage();
+                break;
+            default:
+                throw new IndexOutOfBoundsException();
```
Mick609: I am not sure I understand the purpose of this exception throwing. Why when the command keyword of the userline is not recognizable, the software throws an IndexOutOfBoundsException?

Mick609 (commented):
Clean code with only a few naming convention issues.

PR #108, opened 2021-08-29
[Vanessa Ngoi] iP

ngoivanessa:


src/main/java/Duke.java
```diff
@@ -1,10 +1,131 @@
...
+                String at = input.substring(slash + 4);
+                t = new Event(description, at);
+
+                addTask(t);
+
+            } else { //basic task
+                t = new Task(input);
+
+                addTask(t);
+            }
```
deonchung: Very neat if-else block! Shows good coding standard and quality.

src/main/java/Deadline.java
```diff
@@ -0,0 +1,25 @@
+public class Deadline extends Task {
+
+    /*ATTRIBUTES*/
+
+    private String name; //name of task only (eg return book)
+    protected String by; //due date
```
deonchung: Can rename it to "date" instead of putting a comment to describe the variable

src/main/java/Duke.java
```diff
@@ -1,10 +1,131 @@
...
+        printHorizontalLine();
+    }
+
+    //prints specific task that is done
+    public static void printDoneTask(Task t) {
+        printHorizontalLine();
+        System.out.println("Nice, I've marked this task as done:");
+        String type = t.type;
+        String icon = t.getStatusIcon();
+        System.out.println("[" + type + "]" + " [" + icon + "] " + t.description);
```
deonchung: Put icon and type in the Deadline, Event and ToDo classes. So dont have to repeat getStatusIcon and type multiple times

src/main/java/Event.java
```diff
@@ -0,0 +1,26 @@
+public class Event extends Task {
+
+    /*ATTRIBUTES*/
+
+    private String name; //name of task only (eg return book)
+    protected String at; //time
```
deonchung: Can rename it to "time" instead of putting a comment to describe the variable

src/main/java/Todo.java
```diff
@@ -0,0 +1,17 @@
+public class Todo extends Task {
+
+    /*CONSTRUCTOR*/
+
+    public Todo(String description) {
+        super(description);
+        type = "T";
```
deonchung: Can consider not to declare the type here and just printTaskNotif() function to print the type out

src/main/java/Duke.java
```diff
@@ -1,23 +1,31 @@
...
+
+
+    /*METHODS*/
 
     //adds task to the task list
-    public static void addTask(Task t, int count) {
+    public static void addTask(Task t) {
         printHorizontalLine();
-        tasks[count] = t;
+        tasks[taskCount] = t;
```
WaiKit-nus: I would recommend a different variable name as compared to just a letter t, so it is more easily relatable.

src/main/java/Duke.java
```diff
@@ -1,23 +1,31 @@
...
         t.printTaskNotif();
-        System.out.println("Now you have " + (count + 1) + " tasks in the list");
+        System.out.println("Now you have " + (taskCount + 1) + " tasks in the list");
         printHorizontalLine();
+
```
WaiKit-nus: Try not to leave an empty space inbetween the lines.

src/main/java/Duke.java
```diff
@@ -47,71 +55,77 @@ public static void printHorizontalLine() {
...
-
-    public static void main(String[] args) {
-
+    public static void printStartMessage() {
         System.out.println("Hello! I'm Duke");
         System.out.println("What can I do for you?");
         printHorizontalLine();
+    }
+
+    public static void printByeMessage() {
```
WaiKit-nus: I really like the layout especially the indentations. Good work!

WaiKit-nus: Function names are really logical and easily understandable! Good work.

src/main/java/Event.java
```diff
@@ -0,0 +1,19 @@
+public class Event extends Task {
+
+    private String name; //name of task only (eg return book)
+    protected String at; //time
+
+    public Event(String name, String at) {
+        description = name + " (at: " + at + ")"; //name + time
```
thaddyyz: Might "name" be unambiguous and affect future requirements

src/main/java/Deadline.java
```diff
@@ -0,0 +1,19 @@
+public class Deadline extends Task {
+
+    private String name; //name of task only (eg return book)
+    protected String by; //due date
+
+    public Deadline(String name, String by) {
+        description = name + " (by: " + by + ")"; //name + due date
+        this.by = by;
+        type = "D";
```
thaddyyz: "type" can be plural

src/main/java/Duke.java
```diff
@@ -8,7 +8,8 @@
     public static void addTask(Task t, int count) {
         printHorizontalLine();
         tasks[count] = t;
-        System.out.println("added: " + t.getDescription());
+        t.printTaskNotif();
```
thaddyyz: Can rewrite method name to be more easily understood.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,253 @@
...
+        while (s.hasNext()) {
+            String text = s.nextLine();
+            arr = text.split("\\|");
+
+            boolean status = arr[1].trim().equals("D");
+
+            Task t;
+            switch (arr[0].trim()) {
+            case "T":
+                t = new Todo(arr[2].trim());
```
sinteary: If you include all your main logic inside the `switch case` statements, this function is going to be extremely long. I would recommend extracting these parts out into new functions, each handling the specific functionality

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,40 @@
...
+    public boolean isDone;
+
+
+    /*CONSTRUCTOR*/
+
+    public Task() {
+    }
+
+    public Task(String description) {
+
```
sinteary: A bit of unexpected spacing here which is inconsistent with other parts of your code

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,40 @@
...
+    public void setDone(boolean done) {
+        isDone = done;
+    }
+
+    //set icon as "X" for done, " " for not done
+    public String getStatusIcon() {
+        return (isDone ? "X" : " "); // mark done duke.task with X
+    }
+
+    public void printTaskNotif() {
```
sinteary: I was a bit confused initially because I thought this was "task not if". In this case, this isn't really a "notification" as well, more of just a string representation of the task. Maybe a better name would be `printTaskDisplayString`.
Another thing to think about is - should the _Task_ be responsible for printing itself?

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,253 @@
...
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.ArrayList;
+
+public class Duke {
+
+    /*ATTRIBUTES*/
+
+    private static ArrayList<Task> tasks = new ArrayList<>();
+    //private static Task[] tasks = new Task[100]; //array to store task list
```
sinteary: You should remove code that is commented out and not being used

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,253 @@
...
+        tasks.add(t);
+
+        t.printTaskNotif();
+        System.out.println("Now you have " + tasks.size() + " tasks in the list");
+        printHorizontalLine();
+
+        writeToFile();
+    }
+
+    //prints duke.task list when "list" is keyed by user
```
sinteary: This comment could be redundant if you named the function a bit better (eg. `printTaskList`. We try to minimize comments for obvious parts of code

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,253 @@
...
+        fw.close();
+    }
+
+    public static void readFile() throws FileNotFoundException {
+        Scanner s = new Scanner(taskText);
+        String[] arr;
+
+        while (s.hasNext()) {
+            String text = s.nextLine();
+            arr = text.split("\\|");
```
sinteary: `text` and `arr` can be named more specifically

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,253 @@
...
+                String name = fulltext.substring(0, index).trim();
+                String at = fulltext.substring(index + 4, fulltext.length() - 1).trim();
+
+                t = new Event(name, at);
+                t.isDone = status;
+                tasks.add(t);
+                break;
+            }
+            }
+
```
sinteary: Try to always include a `default` branch in any `switch case` to handle the case where the input matches none of the above cases

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,253 @@
...
+        fw.close();
+    }
+
+    public static void clearFile() throws IOException {
+        FileWriter fw = new FileWriter(filePath);
+        fw.write("");
+        fw.close();
+    }
+
+    public static void readFile() throws FileNotFoundException {
```
sinteary: Can be better named since its not only reading the file but also adding the stored tasks inside. eg. `loadTasksFromFile`
The variables inside (eg. `arr`, `text`) can also be better named

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,253 @@
...
+            switch (arr[0].trim()) {
+            case "T":
+                t = new Todo(arr[2].trim());
+                t.isDone = status;
+                tasks.add(t);
+
+
+                break;
+            case "D": {
+                String fulltext = arr[2];
```
sinteary: Doesn't match the naming convention: variables should be in camelCase i.e.`fullText`

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,253 @@
...
+            input = in.nextLine();
+
+            if (input.equalsIgnoreCase(("bye"))) { //end program
+                isBye = true;
+                printByeMessage();
+
+            } else if (input.equalsIgnoreCase("list")) {
+                printList(); //print duke.task list
+
+            } else if (input.contains("done")) {
```
sinteary: This might be a bit buggy: what if I want to add a task "Explore aban**done**d house" or "Make a well-**done** steak"? Similar for all other parsing methods which use *contains* to check

PR #109, opened 2021-08-29
[Leow Yuan Yang] iP

leowyy99:


src/main/java/Duke.java
```diff
@@ -49,52 +49,118 @@ public static void main(String[] args) {
             // Read inputs
             System.out.println("\tCall out a smurf to do a job for you!");
             Scanner in = new Scanner(System.in);
-            String command;
-            command = in.nextLine();
+            String userInput;
```
mohamad-adam8991: Perhaps this variable name could be in plural?

src/main/java/Duke.java
```diff
@@ -49,52 +49,118 @@ public static void main(String[] args) {
...
+            userInput = in.nextLine();
+            String[] userInputSubstrings = userInput.strip().split(" ", 2);
+            String command = userInputSubstrings[0];
```
mohamad-adam8991: Perhaps a more meaningful name could be used for this?

src/main/java/Duke.java
```diff
@@ -49,52 +49,118 @@ public static void main(String[] args) {
...
-            if (command.startsWith("done")) {
-                // Find item number and update item status
-                int itemNum = Integer.parseInt(command.replaceAll("[^0-9]", "")) - 1;
-                taskList[itemNum].setDone();
+            switch (command) {
+            case "bye":
+                // Exit bot
+                isDone = true;
 
                 // Print output
```
mohamad-adam8991: The format of the comments are well done

src/main/java/Duke.java
```diff
@@ -49,52 +49,118 @@ public static void main(String[] args) {
...
+                } catch (Exception exception) {
+                    System.out.println(SEPARATOR);
+                    System.out.println("\tOops I can't find a Smurf to help you with your task.");
+                    System.out.println(SEPARATOR);
+                }
+                break;
+            case "deadline":
+                try {
+                    // Add task
+                    String deadlineDescription = userInputSubstrings[1]
```
mohamad-adam8991: I like how u broke down the line to keep it under the character limit

src/main/java/Duke.java
leowyy99: Thanks for pointing this out. I thought about it, everytime we pass through the loop, we are only taking 1 input. The whole string is considered userInput. Hence the singular form.

src/main/java/Duke.java
leowyy99: Appreciate that

src/main/java/Duke.java
leowyy99: Do you have any suggestions?

src/main/java/Duke.java
```diff
@@ -1,10 +1,159 @@
...
+
     public static void main(String[] args) {
-        String logo = " ____        _        \n"
-                + "|  _ \\ _   _| | _____ \n"
-                + "| | | | | | | |/ / _ \\\n"
-                + "| |_| | |_| |   <  __/\n"
-                + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
+        boolean isDone = false;
+        Task[] taskList = new Task[100];
```
vincentlauhl: Perhaps this should be the plural of the object (eg : tasks), in accordance to the coding standards for this course.

src/main/java/Duke.java
```diff
@@ -1,10 +1,159 @@
...
+                }
+                break;
+            case "event":
+                try {
+                    String eventDescription = userInputSubstrings[1]
+                            .substring(0, userInputSubstrings[1].indexOf("/"))
+                            .strip();
+                    String eventAt = userInputSubstrings[1].substring((userInputSubstrings[1]
+                                    .indexOf("/") + 3))
+                            .strip();
```
vincentlauhl: Perhaps you can add another layer of abstraction to this

src/main/java/Duke.java
```diff
@@ -1,10 +1,159 @@
...
+            switch (command) {
+            case "bye":
+                isDone = true;
+                printExitMessage();
+                break;
+            case "done":
+                try {
+                    int itemNum = Integer.parseInt(userInput.replaceAll("[^0-9]", ""));
+                    taskList[itemNum - 1].setDone();
+                    printDoneTask(taskList[itemNum - 1], itemNum);
```
vincentlauhl: Maybe instead of itemNum, you could use taskNum, since we are talking about tasks.

src/main/java/Duke.java
```diff
@@ -1,10 +1,159 @@
...
+            System.out.println("\t" + taskList[i]);
+        }
+        System.out.println(SEPARATOR);
+    }
+
+    private static void printErrorMessage() {
+        System.out.println(SEPARATOR);
+        System.out.println("\toops i can't find a smurf to help you with your task.");
+        System.out.println(SEPARATOR);
+    }
```
vincentlauhl: I like how you've packed all the error message into this method

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,229 @@
...
+        System.out.println("\tHey you didn't provide a time :(");
+        System.out.println(SEPARATOR);
+    }
+
+    private static void printExitMessage() {
+        System.out.println(SEPARATOR);
+        System.out.println("\toh shucks! Gargamel is here..we gotta hide");
+        System.out.println(SEPARATOR);
+    }
+}
```
Mick609, line 176: The Duke class contains a bit too many different methods. How about separate some of the methods to other classes? For example, you may have a PrintManager class that only manages all the printing methods.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,229 @@
...
+                printExitMessage();
+                break;
+            case "list":
+                printTaskList(taskList);
+                break;
+            case "done":
+                try {
+                    executeDone(userInputs);
+                } catch (Exception exception) {
+                    printErrorMessage();
```
Mick609: Maybe the printed message can contain more information about the error to have a better user experience? For example, include the exception content?

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,229 @@
...
+                break;
+            }
+        } while (!isDone);
+    }
+
+
+
+    private static void executeEvent(String[] userInputs) throws EmptyDescriptionException, EmptyTimeFieldException {
+        if (userInputs.length < 2) {
+                throw new EmptyDescriptionException();
```
Mick609: Please keep the indentation of statements consistent.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,229 @@
...
+            throw new EmptyTimeFieldException();
+        }
+        String eventDescription = userInputs[1]
+                .substring(0, userInputs[1].indexOf("/"))
+                .strip();
+        if (eventDescription.isBlank()) {
+            throw new EmptyDescriptionException();
+        }
+        String eventAt = userInputs[1].substring((userInputs[1]
+                        .indexOf("/") + 3))
```
Mick609: Again, the indentations of a statement in separate lines should be consistent.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,229 @@
...
+                .strip();
+        taskList[Task.getNumOfTasks()] = new Event(eventDescription, eventAt);
```
Mick609: As the taskList is an array with a fixed length, it would be better if a solution could be added to prevent index out of bound error for the array here.

Mick609 (commented):
Code is easy to understand with minor coding standard issues.

PR #111, opened 2021-08-29
[Andrew Tan Kai Hao] ip

andrewtkh1:


src/main/java/tan/TaskManager.java
```diff
@@ -0,0 +1,223 @@
...
+
+    /**
+     * Returns the Task which is at
+     * the index, null otherwise.
+     *
+     * @param indexTask index of the task to obtain.
+     * @return Task at specified index.
+     */
+    private static Task getTask(int indexTask) {
+        if (indexTask > listOfTasks.size() || indexTask < 0) {
```
snuckerzlol: This could be simplified, maybe bring the expression outside the if

src/main/java/tan/Event.java
```diff
@@ -0,0 +1,43 @@
...
+        super(description);
+        this.dateTime = dateTime;
+    }
+
+    /**
+     * Returns the icon for Event tasks.
+     *
+     * @return Returns the String "E"
+     */
+    public String getIcon() {
```
snuckerzlol: Rather unnecessary method

src/main/java/tan/TaskManager.java
```diff
@@ -0,0 +1,223 @@
...
+                return;
+            }
+            curTask = new ToDo(todoDesc);
+            listOfTasks.add(curTask);
+            break;
+        case "deadline":
+            String deadlineDesc = getDescriptionOfDeadline(userInput);
+            String deadlineDateTime = getDateTimeOfDeadline(userInput);
+            curTask = new Deadline(deadlineDesc, deadlineDateTime);
+            listOfTasks.add(curTask);
```
JoshHDs: Perhaps you could extract out common code between all your case blocks to simplify your code

src/main/java/tan/Deadline.java
```diff
@@ -0,0 +1,42 @@
...
+        super(description);
+        this.dateTime = dateTime;
+    }
+
+    /**
+     * Returns the icon for Deadline tasks.
+     *
+     * @return Returns the String "D"
+     */
+    public String getIcon() {
```
JoshHDs: This method seems rather unnecessary. Perhaps you do not need to have a method to get "D"?

src/main/java/tan/ToDo.java
```diff
@@ -0,0 +1,28 @@
...
+    public ToDo(String description) {
+        super.setTaskName(description);
+    }
+
+    /**
+     * Returns the icon for todo tasks.
+     *
+     * @return Returns the String "T"
+     */
+    public String getIcon() {
```
JoshHDs: It seems rather unnecessary to have a method to return "T", perhaps there is simpler way of doing this?

src/main/java/tan/TaskManager.java
```diff
@@ -0,0 +1,223 @@
+package tan;
+
+import java.util.LinkedList;
+import java.util.List;
+
+public class TaskManager {
+    private static List<Task> listOfTasks = new LinkedList<Task>();
```
JoshHDs: Wow! Seems like an overkill! However, great that you remember how to implement!

src/main/java/tan/TaskManager.java
andrewtkh1: What do you mean by bringing the expression outside the if?

src/main/java/tan/Parser.java
```diff
@@ -0,0 +1,264 @@
...
+import tan.exceptions.DukeFormatExceptions;
+
+import java.time.LocalDate;
+import java.time.format.DateTimeParseException;
+import java.util.regex.PatternSyntaxException;
+
+public class Parser {
+
+    /**
+     * Returns the command of the user's input.
```
stephlewyh, line 16: Good job with the attempt on the javadoc comments! You could also try to shorten your method description - E.g. `splits along spaces and returns` seems to be too much detail.

src/main/java/tan/Parser.java
```diff
@@ -0,0 +1,264 @@
...
+            //+3 to the index as we don't want to capture "/by".
+            String dateTime = x.substring(indexOfSlash + 3);
+            return dateTime.trim();
+        }
+        throw new DukeFormatExceptions("Code could not find '/by'");
+    }
+
+    /**
+     * Returns the description of a deadline task from the user's input.
+     * Else throws a DukeFormatExceptions error
```
stephlewyh: Might be overdoing the descriptions here by including  what exceptions are being thrown. It would suffice to tell what the main function of the method is.

src/main/java/tan/Storage.java
```diff
@@ -0,0 +1,327 @@
...
+            newTask = new ToDo(name, isDone);
+            break;
+        case 1:
+            //Task type Deadline.
+            newTask = new Deadline(name, isDone, date);
+            break;
+        case 2:
+            //Task type event.
+            newTask = new Event(name, isDone, date);
+            break;
```
stephlewyh: How about directly assigning [names](https://nus-cs2113-ay2122s1.github.io/website/se-book-adapted/chapters/codeQuality.html#guideline-name-well) to the taskType? Then these comments would not be necessary.

src/main/java/tan/Storage.java
```diff
@@ -0,0 +1,327 @@
...
+     * Returns the integer corresponding to its task type, else -1.
+     * This function takes in a task-type in String
+     * and returns its corresponding integer value.
+     * Its corresponding task type to integers are,
+     * 0 - todo, 1 - deadline, 2 - event, -1 - unknown.
+     *
+     * @param taskType The Task type in String.
+     * @return The value corresponding to its task type in integer. -1 if unknown.
+     */
+    public static int getTaskTypeInInt(String taskType) {
```
stephlewyh: You have mapped indices to the various taskTypes - this may not be necessary. A [simpler](https://nus-cs2113-ay2122s1.github.io/website/se-book-adapted/chapters/codeQuality.html#practice-kissing) way is use the names directly.

src/main/java/tan/tasktype/ToDo.java
```diff
@@ -0,0 +1,68 @@
+package tan.tasktype;
+
+public class ToDo extends Task {
```
stephlewyh, line 3: This class is easy to read with the javadocs. Good job!

src/main/java/tan/Ui.java
```diff
@@ -0,0 +1,63 @@
...
+
+public class Ui {
+    final static String BORDER = "------------------------------------------------------------------------";
+    private static final Scanner SC = new Scanner(System.in);
+
+    /**
+     * Prints the intro message & border once.
+     */
+    public static void printIntro() {
+        System.out.println("Top of the morning my good sir, what can I do for you on this fine day?");
```
stephlewyh: Note that this may exceed the [120 character limit per line](https://se-education.org/guides/conventions/java/basic.html#layout). For long lines, you might want split them into two separate ones.

src/main/java/tan/TaskList.java
```diff
@@ -0,0 +1,289 @@
...
+            curTask = createDeadlineTask(userInput);
+            if (curTask == null) {
+                return;
+            }
+            break;
+        case "event":
+            curTask = createEventTask(userInput);
+            if (curTask == null) {
+                return;
+            }
```
stephlewyh: There are repeated `curTask == null`-  how about checking `curTask` at the end of the switch statement? This is to improve the[ execution flow](https://nus-cs2113-ay2122s1.github.io/website/se-book-adapted/chapters/codeQuality.html#make-the-happy-path-prominent)

andrewtkh1: Thank you for pointing that out! I didn't notice it. I've changed it.

src/main/java/tan/Storage.java
andrewtkh1: Thank you for the advice! I initially thought of using enums to do this, but thought it might pose an issue when trying to save/load from the data file. Never though of just using the string directly. I have since changed it, Thank you!!!!

PR #114, opened 2021-08-30
[Alwin Ang Yong Siang] iP

alwinangys:


src/main/java/Duke.java
```diff
@@ -1,10 +1,163 @@
...
+                requestList(tasks);
+            } else if (line.contains("done")) {
+                inputDone(line, tasks);
+            } else if (line.contains("undo")) {
+                undoDone(line, tasks);
+            } else if (isValidTaskInput(line)) {
+                taskManager(line, tasks);
+            } else {
+                inputErrorMessage();
+            }
```
brendanlsz: Perhaps you could consider using switch statements to make your code neater?

src/main/java/Duke.java
```diff
@@ -1,10 +1,163 @@
...
+    public static void inputErrorMessage() {
+        printHorizontalLine();
+        System.out.println("I don't understand that. Please try again!");
+        printHorizontalLine();
+    }
+
+    public static void inputManager() {
+        String line;
+        boolean isBye = false;
+        Task[] tasks = new Task[100];
```
brendanlsz: Perhaps you could replace the magic number (100) with a named constant? This will help readers understand your code better.

src/main/java/Duke.java
```diff
@@ -1,10 +1,163 @@
...
+            } else if (isValidTaskInput(line)) {
+                taskManager(line, tasks);
+            } else {
+                inputErrorMessage();
+            }
+        }
+    }
+
+    public static boolean isValidTaskInput(String input) {
+        if (input.toLowerCase().contains("todo") || input.toLowerCase().contains("deadline") || input.toLowerCase().contains("event")) {
```
brendanlsz: Complicated expressions like this can be difficult to read and understand. Maybe you can consider using a few boolean variables here?

src/main/java/Duke.java
```diff
@@ -1,10 +1,163 @@
...
+            return;
+        }
+        System.out.println("Here are the tasks in your list:");
+        for (int i = 1; i < taskCount; i++) {
+            System.out.println(i + "." + tasks[i]);
+        }
+        printHorizontalLine();
+    }
+
+    public static void inputBye() {
```
brendanlsz: Maybe this method could be named more explicitly such as printFarewellMessage? In my opinion, inputBye does not accurately describe the function of this method. I noticed the same issue in several other places too.

src/main/java/duke/Deadline.java
```diff
@@ -0,0 +1,20 @@
+package duke;
+
+public class Deadline extends Task{
+
+    protected String by;
```
Ashley-Lau: Perhaps private would be better here since no classes extends deadline?

Ashley-Lau: Try to use a more descriptive word instead of `by`.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,405 @@
...
+        printHorizontalLine();
+        System.out.println("I don't understand that. Please try again!");
+        printHorizontalLine();
+    }
+
+
+    public static String getCommand(String userInput) throws InvalidTaskDescriptionException {
+        String[] input = userInput.trim().toLowerCase().split(" ");
+        String command = input[0];
+        if (isValidTaskType(userInput)) {
```
Ashley-Lau: Seems to me that this method is doing more than what the method name suggests (which is to extract the command type). Perhaps it would be better to abstract the checks in a separate method or in the inputManager.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,405 @@
...
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Scanner;
+
+public class Duke {
+
+    public static ArrayList<Task> tasks = new ArrayList<>();
```
Ashley-Lau: Something to ponder about: Is class-level variable and access modifier for `tasks` appropriate in this case? This `tasks` could be augmented by any other classes and your duke bot might not work well as intended.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,405 @@
...
+            }
+            switch (command) {
+            case ("bye"):
+                isBye = true;
+                printFarewellMessage();
+                break;
+            case ("list"):
+                requestList();
+                break;
+            case ("done"):
```
Ashley-Lau: Could perhaps abstract out these variable(`done`) and put it as a static variable so that other methods/classes can utilize it.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,405 @@
...
+    public static void loadTaskFile(String filePath) {
+        try {
+            readTaskFile(filePath);
+        } catch (FileNotFoundException e) {
+            System.out.println("File not found");
+        }
+    }
+
+    public static void main(String[] args) {
+        loadTaskFile("data/duke.txt");
```
Ashley-Lau: You could even have a huge try catch block here so that you could not need it in other places (making the code cleaner).

Ashley-Lau (commented):
Good effort in the checks for the inputs!

PR #120, opened 2021-08-30
[Woo Bo Tuan] iP

Bobowoo2468:


CS2113 IP/src/main/java/Duke.java
```diff
@@ -0,0 +1,108 @@
...
+                + "| |_| | |_| |   <  __/\n"
+                + "|____/ \\__,_|_|\\_\\___|\n";
+        System.out.println(logo);
+        String horizontalLine = "________________________";
+        System.out.println("Hello! I'm Duke");
+        System.out.println("What can I do for you?");
+        System.out.println(horizontalLine);
+    }
+
+    public static void listOperations() {
```
rachelkeh: maybe you could consider making this method slightly shorter? maybe in this case, you could directly compare whether the string is "bye", "list" etc instead of creating a boolean variable for all of them?

CS2113 IP/src/main/java/Duke.java
```diff
@@ -0,0 +1,108 @@
...
+            String userInput = sc.nextLine();
+            isBye = userInput.equals("bye");
+            isList = userInput.equals("list");
+            isDone = userInput.startsWith("done");
+            isTodo = userInput.startsWith("todo");
+            isDeadline = userInput.startsWith("deadline");
+            isEvent = userInput.startsWith("event");
+            System.out.println(horizontalLine);
+
+            if (isBye) {
```
rachelkeh: i like how there isn't the arrowhead if block here

CS2113 IP/src/main/java/Duke.java
```diff
@@ -0,0 +1,108 @@
...
+                addTask(taskList, Task.taskCount, userInput, TaskType.EVENT);
+            } else {
+                System.out.println("Error input..");
+            }
+            System.out.println(horizontalLine);
+
+        } while (!isBye);
+    }
+
+    private static void addTask(Task[] taskList, int taskCount, String userInput, TaskType specificTask) {
```
rachelkeh: perhaps you could consider putting this function into another class? because it might be more suitable elsewhere like say the Task class and it'll declutter your main class as well

CS2113 IP/src/main/java/Duke.java
```diff
@@ -0,0 +1,108 @@
...
+        taskList[taskCount] = newTask;
+        Task.taskCount++;
+
+        String printTask = String.format(" [%s][ ] %s", newTask.taskType, newTask.description);
+        String printTaskNumber = String.format("Now you have %d items in the list.", Task.taskCount);
+        System.out.println(printTask);
+        System.out.println(printTaskNumber);
+    }
+
+    private static void markTask(Task[] taskList, String userInput) {
```
rachelkeh: same suggestion as above

CS2113 IP/src/main/java/Duke.java
```diff
@@ -0,0 +1,108 @@
...
+        char userInputIntChar = userInput.charAt(userInput.length() - 1);
+        int userInputInt = Character.getNumericValue(userInputIntChar);
+        taskList[userInputInt].markAsDone();
+        System.out.println("Nice! I've marked this task as done:");
+        String formatOutput = String.format("[%s][%s] %s", taskList[userInputInt].taskType, taskList[userInputInt].getStatusIcon(), taskList[userInputInt].description);
+        System.out.println(formatOutput);
+    }
+
+    private static void listTask(Task[] taskList, int taskCount) {
```
rachelkeh: same suggestion as above

CS2113 IP/src/main/java/Duke.java
```diff
@@ -0,0 +1,108 @@
...
+        Scanner sc = new Scanner(System.in);
+        String horizontalLine = "________________________";
+        boolean isBye;
+        boolean isList;
+        boolean isDone;
+        boolean isTodo;
+        boolean isDeadline;
+        boolean isEvent;
+
+        Task[] taskList = new Task[100];
```
weidak: Not sure if this name would be allowed but plural form should be used on names representing a collection of objects! So maybe it can be "tasks" instead of "taskList"?

CS2113 IP/src/main/java/Duke.java
```diff
@@ -0,0 +1,108 @@
...
+        System.out.println(printTask);
+        System.out.println(printTaskNumber);
+    }
+
+    private static void markTask(Task[] taskList, String userInput) {
+        char userInputIntChar = userInput.charAt(userInput.length() - 1);
+        int userInputInt = Character.getNumericValue(userInputIntChar);
+        taskList[userInputInt].markAsDone();
+        System.out.println("Nice! I've marked this task as done:");
+        String formatOutput = String.format("[%s][%s] %s", taskList[userInputInt].taskType, taskList[userInputInt].getStatusIcon(), taskList[userInputInt].description);
```
weidak: I like how the inputs to the string is so modular. However, if I am not wrong, this line exceeds 120 characters? I think if it is possible you should break it into a few lines.

CS2113 IP/src/main/java/Duke.java
```diff
@@ -0,0 +1,108 @@
...
+        System.out.println(formatOutput);
+    }
+
+    private static void listTask(Task[] taskList, int taskCount) {
+        System.out.println("Here are the tasks in your list:");
+        for (int i = 0; i < taskCount; i++) {
+            int indexNumber = i + 1;
+            String formatOutput = String.format("%d.[%s][%s] %s", indexNumber, taskList[i].taskType, taskList[i].getStatusIcon(), taskList[i].description);
```
weidak: Same as line 91, I think this line exceeds the character count of 120. Maybe this could use a few lines too for code readability.

weidak (commented):
The public functions and classes do not have header comments yet. Otherwise, the code looks mostly free of coding violations and is easy to understand.

CS2113 IP/src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,306 @@
...
+            isBye = userInput.equals("bye");
+            isList = userInput.equals("list");
+            isDone = userInput.startsWith("done");
+            isTodo = userInput.startsWith("todo");
+            isDeadline = userInput.startsWith("deadline");
+            isEvent = userInput.startsWith("event");
+            isDelete = userInput.startsWith("delete");
+            System.out.println(horizontalLine);
+
+            if (isBye) {
```
Leeyp: Well done, do consider using a switch statement in place of "else if" statements as it is slightly preferred (no penalty).

CS2113 IP/src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,306 @@
...
+
+    private static boolean isEmptyDescription(String userInput) {
+        String[] trimDescription = userInput.trim().split("\\s+", 2);
+        return trimDescription.length < 2;
+    }
+
+    private static boolean isIncorrectFormat(String userInput, TaskType specificTask) {
+        switch (specificTask) {
+        case EVENT:
+            boolean hasIncorrectEventPlaceholder = !userInput.contains("/at");
```
Leeyp: Consider abstracting out these check for further clarity.

CS2113 IP/src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,306 @@
...
+        } else{
+            System.out.println("Duke-data directory exists >>::))");
+        }
+    }
+
+    public static void main(String[] args) throws IOException {
+        createDirectory();
+        createFile();
+        setUpDuke();
+        Greet();
```
Leeyp: This method name is not like the others (Capitalization).

Leeyp (commented):
Generally very well done. Coding standards and quality followed well. 

I have left some minor feedbacks for your consideration. No further issues.

PR #125, opened 2021-08-30
[Aditi Chadha] iP

aditichadha1310:
Pull Prequest-1


src/main/java/Duke.java
```diff
@@ -78,6 +118,11 @@ private static void addTaskToList(String userInput) {
...
     }
 
+    /**
+     * Updates the status of the task by marking it as done in the task list.
+     *
+     * @param taskNumberCompleted TaskNumberCompleted stores the task number which has been completed by the user.
+     */
     private static void markTaskAsDone(String userInput) {
         printLine();
         int taskNumberCompleted = Integer.parseInt(userInput.substring(userInput.indexOf(" ") + 1));
```
yanjia1777: Nested methods causes line to be too long. Can add in temporary variables to split it into multiple lines.

src/main/java/Duke.java
```diff
@@ -2,26 +2,52 @@
...
         System.out.println("What can I do for you?");
         printLine();
     }
 
+    /**
+     * Prints a line on the screen
+     */
     public static void printLine() {
         System.out.println("____________________________________________________________");
     }
```
yanjia1777: Good that you use a method to print line instead of repeatedly print it out

src/main/java/Duke.java
```diff
@@ -2,26 +2,52 @@
...
+     */
     public static void main(String[] args) {
         greet();
         runDuke();
         greetBye();
     }
 
+    /**
+     * Greets the user by printing some introductory messages
+     */
```
yanjia1777: Avoid commenting on obvious things

yanjia1777 (commented):
Mostly very good coding standards. However comments may sometimes be excessive especially for obvious things. Makes the code look cluttered even when its very short.

src/main/java/Duke.java
```diff
@@ -1,10 +1,162 @@
...
+        } else if (userInput.startsWith("deadline")) {
+            index = userInput.indexOf("/");
+            scheduledTasks[taskCounter] = new Deadline(userInput.substring(0, index), userInput.substring(index + 3));
+        } else if (userInput.startsWith("event")) {
+            index = userInput.indexOf("/");
+            scheduledTasks[taskCounter] = new Event(userInput.substring(0, index), userInput.substring(index + 3));
+        } else {
+            isTaskValid = false;
+            System.out.println("Invalid task creation statement");
+        }
```
nonrnp: I think it's better to use `switch` statement here

src/main/java/Duke.java
```diff
@@ -1,10 +1,162 @@
...
+    }
+
+    /**
+     * Lists all the tasks in the task list along with their task completion status.
+     * The tasks are enlisted which reveal if they are a "todo", "event" or a "deadline".
+     *
+     * @param taskCompletionStatus TaskCompletionStatus stores true if the task is completed, false otherwise.
+     */
+    private static void List() {
```
nonrnp: method name should starts with lower case

src/main/java/Duke.java
```diff
@@ -1,10 +1,162 @@
...
+        String taskDueBy;
+        String taskDueAt;
+        int index;
+        boolean isTaskValid = true;
+
+        if (userInput.startsWith("todo")) {
+            scheduledTasks[taskCounter] = new Todo(userInput);
+        } else if (userInput.startsWith("deadline")) {
+            index = userInput.indexOf("/");
+            scheduledTasks[taskCounter] = new Deadline(userInput.substring(0, index), userInput.substring(index + 3));
```
nonrnp: You should use named constants instead of random numbers (eg. +3 at the end)

src/main/java/Duke.java
```diff
@@ -1,10 +1,162 @@
...
+     * If user input is "list", it calls the list() function to list all the tasks.
+     * If user input is done x, where x is a valid task number, it calls the MarkTaskAsDone() function to mark the task as done.
+     * If user has scheduled a task by writing a statement beginning with "event", "deadline" or "todo",
+     * then it adds the task in the tasks list by calling the addTaskToList() function.
+     *
+     * @param userInput            userInput stores the input String entered by the user.
+     * @param taskCompletionStatus taskCompletionStatus Stores the status of the task, true if completed, false otherwise.
+     */
+    private static void runDuke() {
+        int i;
```
nonrnp: You should change the variable name to be more descriptive

nonrnp (commented):
Good job for the overall code quality. I only see some minor errors that should be adjusted.

src/main/java/Duke.java
```diff
@@ -1,10 +1,162 @@
...
+
+    /**
+     * Number of tasks scheduled
+     */
+    public static int taskCounter = 0;
+
+    /**
+     * Task type 100 objects to store the tasks the user will create
+     */
+    public static Task[] scheduledTasks = new Task[100];
```
alwinangys: Could replace the magic number (100) with a named constant.

src/main/java/Duke.java
```diff
@@ -1,10 +1,162 @@
+import java.util.Scanner;
+
 public class Duke {
+
+    /**
+     * Number of tasks scheduled
```
alwinangys: taskCounter is quite self-explanatory, the comment is probably trivial in this case.

src/main/java/Duke.java
```diff
@@ -1,10 +1,162 @@
...
+     */
+    public static int taskCounter = 0;
+
+    /**
+     * Task type 100 objects to store the tasks the user will create
+     */
+    public static Task[] scheduledTasks = new Task[100];
+
+    /**
+     * This is the main function which is responsible for executing all the functions
```
alwinangys: Comment is trivial given that this is the expected behaviour of main.

src/main/java/Duke.java
```diff
@@ -1,10 +1,162 @@
...
     public static void main(String[] args) {
-        String logo = " ____        _        \n"
-                + "|  _ \\ _   _| | _____ \n"
-                + "| | | | | | | |/ / _ \\\n"
-                + "| |_| | |_| |   <  __/\n"
-                + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
+        greet();
+        runDuke();
+        greetBye();
```
alwinangys: Good demonstration of SLAP!

src/main/java/Duke.java
alwinangys: Looks like this local variable `i` is never used in this function, perhaps you could consider removing it.

src/main/java/Duke.java
```diff
@@ -1,10 +1,162 @@
...
+            } else {
+                addTaskToList(userInput);
+            }
+            printLine();
+            userInput = in.nextLine();
+        }
+    }
+
+    /**
+     * Greets the user goodbye and the code finishes its execution.
```
alwinangys: While the comment is appropriate, the contents of `greetBye()` does not reflect the code finishing its execution, so maybe that part could be removed.

src/main/java/Duke.java
```diff
@@ -1,10 +1,162 @@
...
+     * @param taskDueBy       TaskDueBy stores the deadline of the task which is created as a deadline.
+     * @param taskDueAt       TaskDueAt stores the event timing of the task created as an event.
+     * @param isTaskValid     IsTaskValid stores true if the task statement emtered by the user is a valid task creation statement and fasle, otherwise.
+     * @param index           Index stores the index of the "/" in the entered String
+     */
+    private static void addTaskToList(String userInput) {
+
+        String taskDescription;
+        String taskDueBy;
+        String taskDueAt;
```
alwinangys: Perhaps for events, taskDueAt may not be the most intuitive name, maybe I could suggest taskDuration? That's if it helps you to understand it better, if not it is fine too.

src/main/java/Duke.java
```diff
@@ -1,10 +1,162 @@
...
+    }
+
+    /**
+     * Lists all the tasks in the task list along with their task completion status.
+     * The tasks are enlisted which reveal if they are a "todo", "event" or a "deadline".
+     *
+     * @param taskCompletionStatus TaskCompletionStatus stores true if the task is completed, false otherwise.
+     */
+    private static void List() {
+        int i;
```
alwinangys: As the integer `i` is only used in the for loop in this function, this variable should be defined in the least possible scope (only define `i` in the for loop)

src/main/java/Event.java
```diff
@@ -0,0 +1,28 @@
...
+    }
+
+    /**
+     * returns the string in a particular format.
+     *
+     * @return Task description and event time in a particular format.
+     */
+    @Override
+    public String toString() {
+        return ("[E]" + super.toString() + " (at:" + this.at + ")");
```
alwinangys: There is no need for `this` as there are no parameters in this function that shadows the instance variable.

src/main/java/Task.java
```diff
@@ -0,0 +1,61 @@
...
+
+    /**
+     * Returns the task status in a particular String format.
+     *
+     * @return Task completion Status in a String format.
+     */
+    @Override
+    public String toString() {
+        String taskStatus = isDone ? "X" : " ";
+        return ("[" + taskStatus + "] " + this.description);
```
alwinangys: Similar issue to Event class

src/main/java/Deadline.java
```diff
@@ -0,0 +1,28 @@
...
+    }
+
+    /**
+     * returns the string in a particular format.
+     *
+     * @return Task description and deadline in a particular format.
+     */
+    @Override
+    public String toString() {
+        return ("[D]" + super.toString() + " (by:" + this.by + ")");
```
alwinangys: Similar issue to Event class

alwinangys (commented):
Overall good demonstration of OOP and abstraction. However there are some inappropriately declared variables and trivial comments, which could make the code look unnecessarily long. Good work!

src/main/java/Duke.java
aditichadha1310: Sure, I will do that.

src/main/java/Duke.java
aditichadha1310: Okay sure, I will eliminate them.

src/main/java/Duke.java
aditichadha1310: Oh Okay

src/main/java/Duke.java
aditichadha1310: Thank you:)

src/main/java/Duke.java
aditichadha1310: Sure

src/main/java/Duke.java
aditichadha1310: Sure I will do that.

aditichadha1310: Sure I will do it. Thanks for notifying.

src/main/java/Duke.java
aditichadha1310: Since greetbye() is the last function that runs in main(), as soon as it finishes its execution the code would have completed all the steps required for its complete execution. That's why I mentioned it. But I will eliminate it in the next commit.

src/main/java/Duke.java
aditichadha1310: I chose TaskDueAt because we are using the "at" to distinguish an event, so i somehow wanted to use "At" in the string name to make it clear for the readers. But I will definitely consider changing the names.

src/main/java/Duke.java
aditichadha1310: Thanks for notifying. I will do it!

src/main/java/Duke.java
aditichadha1310: Yes sure, I will try implementing it. Thanks for the suggestion.

src/main/java/Duke.java
aditichadha1310: Sure I will do it.

src/main/java/Duke.java
aditichadha1310: Oh I am sorry for this. I will definitely change it. Thanks for notifying.

src/main/java/Duke.java
aditichadha1310: Okay sure.

src/main/java/Event.java
aditichadha1310: Sure , I will change it.

src/main/java/Task.java
aditichadha1310: Yeah, I will change it.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,227 @@
...
+import exception.DukeException;
+
+import java.util.ArrayList;
+import java.util.Scanner;
+
+public class Duke {
+
+    /**
+     * Number of tasks scheduled
+     */
```
EthanWong2212: Since the variable name and code is self explanatory perhaps it would be better to refrain from repeating description in comment

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,227 @@
...
+    }
+
+    /**
+     * Greets the user by printing some introductory messages
+     */
+    private static void greet() {
+        printLine();
+        System.out.println("Hello! I'm Duke");
+        System.out.println("What can I do for you?");
+        printLine();
```
EthanWong2212: Maybe it would be better to group the strings and use a constant to store them to avoid use of magic literals (strings in this case)

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,227 @@
...
+        }
+
+        if (isTaskValid) {
+            printLine();
+            System.out.println("Got it. I've added this task:");
+            System.out.println(" " + scheduledTasks.get(scheduledTasks.size() - 1));
+            taskCounter++;
+            System.out.println("Now you have " + scheduledTasks.size() + " tasks in the list.");
+        }
+    }
```
EthanWong2212: Method is very long. Maybe it will be better to abstract out portions of the code (e.g. validity check) to improve quality and readability.

src/main/java/duke/Deadline.java
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    /**
+     * returns the string in a particular format.
+     *
+     * @return Task description and deadline in a particular format.
+     */
+    @Override
+    public String toString() {
+        return ("[D]" + super.toString() + " (by:" + this.by + ")");
```
EthanWong2212, line 28: Good that you use super (parent class) to reuse portions of the code

src/main/java/duke/TaskType.java
```diff
@@ -0,0 +1,5 @@
+package duke;
+
+public enum TaskType {
+    NONE, TODO, DEADLINE, EVENT
+}
```
EthanWong2212, line 8: Good use of enum

src/main/java/duke/Event.java
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    /**
+     * returns the string in a particular format.
+     *
+     * @return Task description and event time in a particular format.
+     */
+    @Override
+    public String toString() {
+        return ("[E]" + super.toString() + " (at:" + this.at + ")");
```
EthanWong2212, line 28: Perhaps you can use an enum for the task type icons to prevent use of magic strings?

EthanWong2212 (commented):
Good Job overall, would be good to remove use of magic literals. Keep up the good work!

src/main/java/duke/Deadline.java
aditichadha1310: Thanks

src/main/java/duke/Duke.java
aditichadha1310: Sure I will do that.

src/main/java/duke/Duke.java
aditichadha1310: Yeah sure I will work on it.

src/main/java/duke/Duke.java
aditichadha1310: Yes definitely. I will work on it.

src/main/java/duke/Event.java
aditichadha1310: That's a very valid suggestion. i will do it.

src/main/java/duke/TaskType.java
aditichadha1310: Thanks.

PR #127, opened 2021-08-30
[Louis] iP

LouisLouis19:
Pull Request 1: Add Level-0, Level-1, Level-2, Level-3, A-Classes, A-CodingStandard, and Refactor code


src/main/java/Duke.java
```diff
@@ -2,11 +2,19 @@
 import java.util.Arrays;
 
 public class Duke {
+    //List of Special User Commands
+    public static final String EXIT_STRING = "bye";
+    public static final String LIST_COMMAND = "list";
+    public static final String DONE_COMMAND = "done";
+    public static final String TODO_COMMAND = "todo";
+    public static final String DEADLINE_COMMAND = "deadline";
+    public static final String EVENT_COMMAND = "event";
```
siyuan-h: It is quite good practice to define the command using constant, which will greatly reduce time needed to modify the code once there is a change to be made.

src/main/java/Duke.java
```diff
@@ -52,7 +52,16 @@ public static void returnException() {
         System.out.println("____________________________________________________________");
     }
 
-    public static void printTaskDone(String[] commands, int index, Task[] taskList) {
+    public static boolean checkValidDoneInstruction(String inWord) {
+        String[] commands = inWord.split(" ");
+        if (commands.length != 2 || !isNumeric(commands[1])) {
+            return false;
+        }
```
siyuan-h: It is good to do validation check to make the program robust.

src/main/java/Duke.java
```diff
@@ -81,34 +177,55 @@ public static void main(String[] args) {
...
-                index += 1;
-                //Add additional task
+                break;
+            default:
+                returnException();
+                break;
             }
             inWord = scan.nextLine();
+            instruction = inWord.split("\\s+", 2);
+            instructionType = instruction[0];
```
siyuan-h: can add more comment to increase the readability.

src/main/java/Duke.java
```diff
@@ -62,17 +71,104 @@ public static void printTaskDone(String[] commands, int index, Task[] taskList)
...
+
+    public static boolean checkValidDeadline(String inWord) {
+        String[] commands = inWord.split("\\s+", 2);
+        if (commands.length != 2 || !inWord.contains(" /by")) {
+            return false;
+        }
+
+        String[] details = commands[1].split("/by", 2);
+        String descriptionDetails = details[0].trim();
+        String descriptionDeadline = details[1].trim();
```
siyuan-h: can change the variable name of descriptionDeadline to increase readability, since people may be confused about the difference between descriptionDeadline and deadline description.

siyuan-h (commented):
Overall this is a great example of code with robustness and high readability.

src/main/java/Duke.java
```diff
@@ -1,10 +1,259 @@
...
+
+    public static void printDukeGreet() {
         String logo = " ____        _        \n"
                 + "|  _ \\ _   _| | _____ \n"
                 + "| | | | | | | |/ / _ \\\n"
                 + "| |_| | |_| |   <  __/\n"
                 + "|____/ \\__,_|_|\\_\\___|\n";
         System.out.println("Hello from\n" + logo);
+
+        //Greeting the User
```
kum-wh: Nice inclusion of comments to describe what the purpose of that code

src/main/java/Duke.java
```diff
@@ -1,10 +1,259 @@
...
+        System.out.println(LINE);
+        System.out.println(" Hello! I'm Duke");
```
kum-wh: can be combined into one print function and format the string instead of calling the same function multiple times

src/main/java/Duke.java
```diff
@@ -1,10 +1,259 @@
...
+            return false;
+        }
+        return true;
+    }
+
+    public static void returnException() {
+        System.out.println(LINE);
+        System.out.println(EXCEPTION);
+        System.out.println(LINE);
+        System.out.println();
```
kum-wh: Can use the System line Separator instead of printing a empty string on a new line, as this can cause some confusion to readers

src/main/java/Duke.java
```diff
@@ -1,10 +1,259 @@
+import java.util.Scanner;
+import java.util.Arrays;
+
 public class Duke {
-    public static void main(String[] args) {
+    //List of Special User Commands
+    public static final String EXIT_STRING = "bye";
```
kum-wh: Good use of using constants preventing the issue of magical numbers or in this case magical strings

kum-wh (commented):
Overall code quality is pretty good.  Code is easily understandable.

src/main/java/Duke.java
LouisLouis19: Thank you for the insightful feedback, I will address this issue in my next code update

src/main/java/Duke.java
LouisLouis19: Thank you for the naming suggestion, will address this issue in my next update!

src/main/java/Duke.java
LouisLouis19: isn't this code fragment taken from within a print function? Is it necessary to refactor these print statements within a print function into another separate print function?

src/main/java/Duke.java
LouisLouis19: Thank you for the insightful feedback, will address them in my next update

src/main/java/Duke/RunFile/DukeProgram.java
```diff
@@ -0,0 +1,359 @@
...
+
+    public static void manageDoneInstruction(String inWord, ArrayList<Task> taskList) {
+        try {
+            printTaskDone(inWord, taskList);
+        } catch (DukeException invalidDoneException){
+            DukeException.invalidDoneException();
+        }
+    }
+
+    public static boolean checkValidEvent(String inWord) {
```
OngDeZhi: Boolean variables/methods should be named to sound like booleans, so you may want to use prefixes like is, has, was, etc.

src/main/java/Duke/RunFile/DukeProgram.java
```diff
@@ -0,0 +1,359 @@
...
+        taskList.add(newItem);
+        System.out.println(LINE);
+        System.out.println(" Got it. I've added this task:");
+        System.out.println("   " + newItem);
+        System.out.println(" Now you have " + taskList.size() +" tasks in the list.");
+        System.out.println(LINE);
+        System.out.print(System.lineSeparator());
+    }
+
+    public static void manageEvent(String inWord, ArrayList<Task> taskList) {
```
OngDeZhi: You may want to consider renaming the method since it is just calling the print method and not really managing the events.

src/main/java/Duke/SaveFile/DataSaver.java
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    public static void addToTaskList(ArrayList<Task> taskList, String newTask) throws DukeException {
+        String[] taskDetails = newTask.split("\\|");
+
+        if (!checkValidDetails(taskDetails)) {
+            throw new DukeException();
+        }
+
+        String type = taskDetails[0].trim();
```
OngDeZhi, line 84: Could name it better like taskType.

src/main/java/Duke/Exception/DukeException.java
```diff
@@ -0,0 +1,75 @@
+package Duke.Exception;
```
OngDeZhi, line 1: Names representing packages should be in all lower case.

src/main/java/Duke/SaveFile/DataSaver.java
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    public static void parseType(Task task, StringBuilder parsedTask) {
+        String type = task.getType();
+        parsedTask.append(type).append(DIVIDER);
+    }
+
+    public static void parseStatus(Task task, StringBuilder parsedTask) {
+        if (task.getIsDone()) {
+            parsedTask.append("1").append(DIVIDER);
```
OngDeZhi, line 191: You may want to define what "1" and "0" means, if it refers to done status, then do remember to use the constant you have defined.

src/main/java/Duke/RunFile/DukeProgram.java
```diff
@@ -0,0 +1,359 @@
...
+        return true;
+    }
+
+    public static boolean isValidDoneInstruction(String inWord, ArrayList<Task> taskList) {
+        if (!inWord.contains(" ")) {
+            return false;
+        }
+
+        String[] commands = inWord.split(" ");
+        if (commands.length != 2) {
```
OngDeZhi: Avoid magic numbers - extract them out and name it.

src/main/java/Duke/Exception/DukeException.java
```diff
@@ -0,0 +1,75 @@
...
+
+import java.io.IOException;
+
+public class DukeException extends Exception{
+    public static final String LINE = "____________________________________________________________";
+
+    public static void emptyTodoException() {
```
OngDeZhi, line 15: Notice that for the methods you have here, you are repeating the same few print statement, so you may want to extract them out and create a new method.

PR #136, opened 2021-08-30
[Richard Willie] iP

richwill28:
## Changelog

* **Added `list` feature**
User can now displays the list of tasks.
* **Added `done` feature**
User can now marks task as done.
* **Minor customizations**
Changed some commands and logo.
* **New types of tasks**
Added three types of tasks, namely `Todo`, `Deadline`, and `Event`.
* **Automated text UI testing**
Enable automated UI testing by executing `./runtest.sh`. This is useful for regression testing.
* **Improved tasks list implementation**
Tasks list is now implemented using dynamic array and is managed by the `TaskList` class.
* **Refactored UI components**
UI components are refactored to a new `Ui` class.
* **Refactored user responses**
User responses validation are abstracted using the `Command` class.
* **Added `JavaDoc` comments**
To make the codebase readable.


src/main/java/Duke.java
```diff
@@ -1,10 +1,164 @@
...
+
+/**
+ * This class implements the Project Duke, a Personal
+ * Assistant Chatbot that helps a user to keep track
+ * of various things. Currently, Duke can greet user;
+ * add/list user tasks, and mark them as "done".
+ *
+ * @author richwill28
+ */
+class Duke {
```
aditichadha1310: You might want to show the access level of the class Duke?

richwill28: Thanks for pointing this out! I should declare the class as `public` instead of the `default access` to allow other packages to access it.

src/main/java/Command.java
```diff
@@ -0,0 +1,82 @@
...
+ */
+class Command {
+    /** User command */
+    String command;
+
+    /**
+     * The constructor method. Initializes user command.
+     *
+     * @param command User command.
+     */
```
crabnuggets: I think your habit of documentation throughout is great and it helps readers understand the code much better. But perhaps for straightforward methods like constructors, getters and setters, you can consider removing these comments as it can make looking through the entire code a longer process than it should too.

src/main/java/Deadline.java
```diff
@@ -0,0 +1,32 @@
+/**
+ * The Deadline class is a subclass of Task. It
+ * provides an additional parameter, deadline.
+ *
+ * @author richwill28
+ */
+class Deadline extends Task {
+    /** Task deadline */
+    protected String by;
```
crabnuggets: `by` could come off as an ambiguous/awkward sounding field name. Perhaps something clearer like "date" or "time" would be better.

src/main/java/Duke.java
```diff
@@ -1,10 +1,164 @@
...
+                this.markAsDone(response.getCommand());
+            } else if (response.isTodo()) {
+                this.addTodo(response.getCommand());
+            } else if (response.isDeadline()) {
+                this.addDeadline(response.getCommand());
+            } else if (response.isEvent()) {
+                this.addEvent(response.getCommand());
+            } else {
+                Ui.error();
+            }
```
crabnuggets: I like how neat this entire if-else block is and how it follows coding standards and is an example of good code quality.

src/main/java/Event.java
```diff
@@ -0,0 +1,32 @@
+/**
+ * The Event class is a subclass of Task. It
+ * provides an additional parameter, time period.
+ *
+ * @author richwill28
+ */
+class Event extends Task {
+    /** Time period of the task */
+    protected String at;
```
crabnuggets: `at` seems to be another instance of an ambiguous field name.

src/main/java/Ui.java
```diff
@@ -0,0 +1,54 @@
+/**
+ * This class handles the User Interface of Duke.
+ *
+ * @author richwill28
+ */
+class Ui {
+    /** Decorative logo */
+    public static final String LOGO =
```
crabnuggets: As mentioned, it might be worth considering removing comments like these, especially so when the names you have chosen for such class members seem to clearly suggest what they refer to.

crabnuggets (commented):
Looks good and is very nicely documented. 
Provided some minor suggestions with regards to coding standards and quality.

src/main/java/Ui.java
richwill28: Thanks for the suggestion! I agree that this comment introduced unnecessary clutter on my code, will remove it on the next commit.

src/main/java/Duke.java
richwill28: Thanks. 😊

src/main/java/Command.java
richwill28: Noted. On second thought, all the getter and boolean methods that I used in this class are straightforward, and I do agree that they don't need extra documentations.

src/main/java/Deadline.java
richwill28: Noted.

src/main/java/Event.java
richwill28: Noted.

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,107 @@
...
+            }
+            String description = params[2];
+            switch (taskType) {
+            case "T":
+                return new Todo(description, isDone);
+            case "D":
+                String by = params[3];
+                return new Deadline(description, by, isDone);
+            case "E":
+                String at = params[3];
```
nishanthelango: Avoid using magic strings such as "T". You can consider extracting them into constants.

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,107 @@
...
+                return new Event(description, at, isDone);
+            default:
+                throw new IllegalArgumentException("Error: Invalid format. Unable to deserialize string.");
+            }
+        } catch (ArrayIndexOutOfBoundsException e) {
+            throw new IllegalArgumentException("Error: Invalid format. Unable to deserialize string.");
```
nishanthelango: Consider extracting reused strings into constants.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,300 @@
...
+
+    /**
+     * Add Deadline to the list of tasks.
+     *
+     * @param response User response, consists of
+     *                 description and deadline.
+     */
+    public void addDeadline(String response) throws DukeException {
+        String[] params = response.replace("deadline", "").strip().split("/by");
+        if (params.length > 2) {
```
nishanthelango: Consider extracting magic numbers such as 2 into constants.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,300 @@
...
+        sc = new Scanner(System.in);
+        taskList = new TaskList();
+    }
+
+    public String getResponse() {
+        return sc.nextLine();
+    }
+
+    /** List all the tasks. */
+    public void list() {
```
nishanthelango: Consider using more descriptive names for methods to describe what exactly the method is doing. Perhaps "listTasks" might be more appropriate instead of "list"?

src/main/java/duke/task/Task.java
```diff
@@ -0,0 +1,107 @@
...
+        return description;
+    }
+
+    /**
+     * Deserialize a line of string and match to suitable
+     * task format.
+     *
+     * @param line A line of string.
+     * @return The new task after deserialization.
+     */
```
nishanthelango: Consider starting the first sentence of the Javadoc comment with "Deserializes" instead of "Deserialize", as per coding standards.

src/main/java/duke/task/Task.java
richwill28: Noted.

src/main/java/duke/task/Task.java
richwill28: Good idea!

src/main/java/duke/Duke.java
richwill28: Noted.

src/main/java/duke/Duke.java
richwill28: Noted.

src/main/java/duke/task/Task.java
richwill28: Noted.

PR #146, opened 2021-08-30
[Sicheng] iP

Isabella-L:


src/main/java/Duke.java
```diff
@@ -1,10 +1,137 @@
...
+        }
+
+        public String getStatusIcon() {
+            return (isDone ? "X" : " ");
+        }
+        public void markAsDone() {
+            this.isDone = true;
+        }
+
+    }
```
Woolicious98: Perhaps this Task class could be contain as class file names Task?

src/main/java/Duke.java
```diff
@@ -1,10 +1,137 @@
...
+                "                                       \\|                    `.\\\n" +
+                "  S. Lu - I solemnly swear that I am up to no good.\n";
         System.out.println("Hello from\n" + logo);
+        System.out.println("Hello! I'm Duke\n"
+                + "What can I do for you?\n");
+    }
+
+    public static void exit() {
+        System.out.println("Bye. Hope to see you again soon!");
+    }
```
Woolicious98: Maybe function name could be call printExitMessage instead?

src/main/java/Duke.java
```diff
@@ -1,10 +1,137 @@
...
+    //index is the position in string, index+1 is the bullet number
+    public static void add(String command, int index) {
+        commands[index] = new Task(command);
+        commands[index].taskType = getTask(command,index);
+        System.out.println("Got it. I've added this task: \n" +
+                "\t[" + commands[index].taskType + "]" + "[ ] " +
+                commands[index].description + "\n" +
+                "Now you have " + (index+1) + (index == 0?" task":" tasks") + " in the list."
+                );
+        }
```
Woolicious98: I think the function name for the commands could be more specific like addTask? The same goes for the other commands methods.

src/main/java/Duke.java
```diff
@@ -1,10 +1,137 @@
...
+            return true;
+        }
+        return false;
+    }
+
+    public static char getTask(String command, int i) {
+        //.replaceAll method remove word or char from string
+        //to obtain strings after certain words, we can use
+        //StringUtil.substringAfter(String,word)
+        //however, to avoid external library download, i use a more tedious method here
```
Woolicious98: Perhaps comments could be added following the Java coding standard with the /**/ appended on top of these methods.

src/main/java/Duke.java
```diff
@@ -1,10 +1,137 @@
...
+        }
+        public void markAsDone() {
+            this.isDone = true;
+        }
+
+    }
+
+    static String exitTrigger = "bye";
+    static String listTrigger = "list";
+    static String doneTrigger = "done";
```
Woolicious98: Constants name should be in all caps, perhaps you could change them to TRIGGER_EXIT?

Woolicious98 (commented):
Overall great job!

src/main/java/Duke.java
```diff
@@ -1,10 +1,137 @@
...
+    }
+
+    //index is the position in string, index+1 is the bullet number
+    public static void add(String command, int index) {
+        commands[index] = new Task(command);
+        commands[index].taskType = getTask(command,index);
+        System.out.println("Got it. I've added this task: \n" +
+                "\t[" + commands[index].taskType + "]" + "[ ] " +
+                commands[index].description + "\n" +
+                "Now you have " + (index+1) + (index == 0?" task":" tasks") + " in the list."
```
markuslyq: I think adding spaces before and after the plus sign would make things a lot neater?

src/main/java/Duke.java
markuslyq: Perhaps a function name called printExitMessage would be better?
```suggestion
    public static void printExitMessage() {
        System.out.println("Bye. Hope to see you again soon!");
    }
```

src/main/java/Duke.java
markuslyq: I think it would be better if the Task class is in a separate java file called Task.java?

src/main/java/Duke.java
markuslyq: If I am not wrong, constant names should be in caps and underscore as a separator, so maybe a few changes can be made to those names?
```suggestion
    static String EXIT_TRIGGER= "bye";
    static String LIST_TRIGGER = "list";
    static String DONE_TRIGGER = "done";
```

markuslyq (commented):
Overall, although the code works, i think more OOP concepts can be implemented.

src/main/java/Duke.java
mohamedirfansh: Task class could be in a separate file itself since this file already contains the Duke class.

src/main/java/Duke.java
mohamedirfansh: These string constants could be all caps with perhaps a final keyword in front to match coding standards.

src/main/java/Duke.java
```diff
@@ -1,10 +1,137 @@
...
+        }
+
+    }
+
+    static String exitTrigger = "bye";
+    static String listTrigger = "list";
+    static String doneTrigger = "done";
+    static Task[] commands = new Task[100];
+
+    public static void greet() {
```
mohamedirfansh, line 28: You could also name this function `printGreetings`.

src/main/java/Duke.java
```diff
@@ -1,10 +1,137 @@
...
+                "   /_\\  | |           //_____//       .||`  _   `._,' | |   | | \\ `-' /| |\n" +
+                "        /_\\           `------'        \\ |               `.\\  | |  `._,' /_\\\n" +
+                "                                       \\|                    `.\\\n" +
+                "  S. Lu - I solemnly swear that I am up to no good.\n";
         System.out.println("Hello from\n" + logo);
+        System.out.println("Hello! I'm Duke\n"
+                + "What can I do for you?\n");
+    }
+
+    public static void exit() {
```
mohamedirfansh, line 47: You could name this function as `printExitMessage` to improve readability.

src/main/java/Duke.java
```diff
@@ -1,10 +1,137 @@
...
+        commands[index] = new Task(command);
+        commands[index].taskType = getTask(command,index);
+        System.out.println("Got it. I've added this task: \n" +
+                "\t[" + commands[index].taskType + "]" + "[ ] " +
+                commands[index].description + "\n" +
+                "Now you have " + (index+1) + (index == 0?" task":" tasks") + " in the list."
+                );
+        }
+
+    public static void list() {
```
mohamedirfansh, line 62: You could name this method as `printTaskList` to make it a little more descriptive.

src/main/java/Duke.java
```diff
@@ -1,10 +1,137 @@
...
+        //to obtain strings after certain words, we can use
+        //StringUtil.substringAfter(String,word)
+        //however, to avoid external library download, i use a more tedious method here
+        if (command.contains("todo")) {
+            commands[i].description = command.replaceAll("todo","");
+            return 'T';
+        }
+        if (command.contains("deadline")) {
+            String time = command.substring(command.indexOf("/by")+4);
+            String task = command.substring(command.indexOf("deadline")+9,command.indexOf("/by"));
```
mohamedirfansh, line 100: You could try to remove these magic numbers.

src/main/java/Duke.java
```diff
@@ -1,10 +1,137 @@
+import java.util.Scanner;
+
 public class Duke {
-    public static void main(String[] args) {
-        String logo = " ____        _        \n"
-                + "|  _ \\ _   _| | _____ \n"
-                + "| | | | | | | |/ / _ \\\n"
-                + "| |_| | |_| |   <  __/\n"
-                + "|____/ \\__,_|_|\\_\\___|\n";
+    public static class Task {
```
mohamedirfansh: Perhaps, the Task class could be in a separate file since this file already contains the Duke class.

src/main/java/Duke.java
```diff
@@ -1,10 +1,137 @@
...
+        public String getStatusIcon() {
+            return (isDone ? "X" : " ");
+        }
+        public void markAsDone() {
+            this.isDone = true;
+        }
+
+    }
+
+    static String exitTrigger = "bye";
```
mohamedirfansh: You could try to make string constants as all caps.

src/main/java/Duke.java
mohamedirfansh: You could also name this function `printGreetings` for better readability.

src/main/java/Duke.java
mohamedirfansh: You could name this function as `printExitMessage` to improve readability.

src/main/java/Duke.java
mohamedirfansh: You could name this method as `printTaskList` to make it a little more descriptive.

src/main/java/Duke.java
```diff
@@ -1,10 +1,137 @@
...
+        }
+        if (command.contains("deadline")) {
+            String time = command.substring(command.indexOf("/by")+4);
+            String task = command.substring(command.indexOf("deadline")+9,command.indexOf("/by"));
+            commands[i].description = task + "(by: " + time + ")";
+            return 'D';
+        }
+        if (command.contains("event")) {
+            String time = command.substring(command.indexOf("/at")+4);
+            String task = command.substring(command.indexOf("event")+6,command.indexOf("/at"));
```
mohamedirfansh: You could remove these magic numbers by assigning these values to a variable, and using the variable instead.

src/main/java/Duke.java
```diff
@@ -1,10 +1,137 @@
...
+        protected boolean isDone;
+        protected char taskType;
+
+        public Task(String description) {
+            this.description =  description;
+            this.isDone = false;
+        }
+
+        public String getStatusIcon() {
+            return (isDone ? "X" : " ");
```
iamabhishek98: Avoid using magic strings like "X" - extract them out to constants and name them using proper naming conventions.

src/main/java/Duke.java
iamabhishek98: or you could name it `printGreetMessage()`

src/main/java/Duke.java
iamabhishek98: Agreed!

src/main/java/Duke.java
```diff
@@ -1,10 +1,137 @@
...
         System.out.println("Hello from\n" + logo);
+        System.out.println("Hello! I'm Duke\n"
+                + "What can I do for you?\n");
+    }
+
+    public static void exit() {
+        System.out.println("Bye. Hope to see you again soon!");
+    }
+
+    //index is the position in string, index+1 is the bullet number
```
iamabhishek98: Please maintain spacing between the comment specifier (//) and the first word of the comment. Also, try to make your comments more descriptive instead of just explaining what the variables mean

src/main/java/Duke.java
iamabhishek98: Few fundamental issues with this piece of code:
1. Having such complex computational logic in your print statements makes the code extremely long and hard to understand. The sole purpose of a print statement is to print something and nothing else. Therefore, its better if you separate these nested functions to different lines.
2. Also avoid using ternary operators as they can make your code unnecessarily hard to understand. It would be better to change the logic to use if ... else if statements instead and then conditionally print
3. Please maintain spacing between the operators and the operands. Your style of using spacing in code seems to be quite inconsistent

src/main/java/Duke.java
```diff
@@ -1,10 +1,137 @@
...
+            char taskType = commands[i].taskType;
+            System.out.print(bullet_num +". ");
+            System.out.print("[" + taskType +"]" + "["+ statusIcon +"] ");
+            System.out.println(commands[i].description);
+        }
+    }
+
+    public static boolean mark_done(String command){
+        //command will be "done X"
+        String cmd[] = command.split(" ", 2);
```
iamabhishek98: Try to make your variable names more meaningful and representative of what you are trying to do

src/main/java/Duke.java
```diff
@@ -1,10 +1,137 @@
...
+        String firstword = cmd[0];
+        //command.contains(word) is not used to obtain the number behind
```
iamabhishek98: Comments should be used to describe what is going on but that does not seem to be the case here. Avoid having such comments.

src/main/java/Duke.java
iamabhishek98: Agreed!

src/main/java/Duke.java
```diff
@@ -1,10 +1,137 @@
...
+    public static void main(String[] args) {
+        greet();
+
+        Scanner in = new Scanner(System.in);
+        String command = in.nextLine();
+
+        boolean should_exit = command.equals(exitTrigger);
+
+        for(int i = 0; !command.equals(exitTrigger) ; i++) {
+            if (mark_done(command)) {
```
iamabhishek98: Try to be consistent with your naming conventions. It seems like you are alternating between snake case and camel case. For variable and method names, always stick to using camel case.

src/main/java/Duke.java
iamabhishek98: These variables are constants and should follow the proper naming conventions

src/main/java/Duke.java
```diff
@@ -1,10 +1,137 @@
...
+
+    public static void main(String[] args) {
+        greet();
+
+        Scanner in = new Scanner(System.in);
+        String command = in.nextLine();
+
+        boolean should_exit = command.equals(exitTrigger);
+
+        for(int i = 0; !command.equals(exitTrigger) ; i++) {
```
iamabhishek98: This does not seem like an appropriate use of a for loop. Perhaps a while loop is more suitable in this case

src/main/java/Duke.java
iamabhishek98: Agreed!

src/main/java/Duke.java
```diff
@@ -1,10 +1,137 @@
...
+            System.out.print(bullet_num +". ");
+            System.out.print("[" + taskType +"]" + "["+ statusIcon +"] ");
+            System.out.println(commands[i].description);
+        }
+    }
+
+    public static boolean mark_done(String command){
+        //command will be "done X"
+        String cmd[] = command.split(" ", 2);
+        String firstword = cmd[0];
```
iamabhishek98: Variable names which have multiple words should use the camel case naming convention

iamabhishek98 (commented):

PR #151, opened 2021-08-30
[Hu Siyuan] iP

siyuan-h:
A pull request


src/main/java/Duke.java
```diff
@@ -1,10 +1,180 @@
...
+            Event event = new Event(description, time);
+            System.out.println("       " + event.doneToString());
+            break;
+        default:
+            System.out.println("       Unknown Object");
+        }
+        System.out.println("    ____________________________________________________________");
+    }
+
     public static void main(String[] args) {
```
joshualeeky, line 12: I think it may be better to place this function at the top.

src/main/java/Duke.java
```diff
@@ -1,10 +1,180 @@
...
+        Scanner in = new Scanner(System.in);
+        // get user input
+        line = in.nextLine();
+        while (!line.equals("bye")) {
+            if (line.equals("list")) {
+                showList();
+                line = in.nextLine();
+                continue;
+            }
+            if (line.startsWith("done")) {
```
joshualeeky: I think this could be an else if instead of just an if.

src/main/java/Duke.java
```diff
@@ -1,10 +1,180 @@
...
+        }
+        System.out.println("    ____________________________________________________________");
+    }
+
+    private static void addList() {
+        String line;
+        Scanner in = new Scanner(System.in);
+        // get user input
+        line = in.nextLine();
+        while (!line.equals("bye")) {
```
joshualeeky: I think you can consider using functions for each of the if cases to improve the code readability.

src/main/java/Duke.java
```diff
@@ -1,10 +1,180 @@
...
+                time = line.substring(seperate + 4);
+                Event event = new Event(description, time);
+                list.add(description);
+                date.add(time);
+                System.out.println("       " + event);
+            } else {
+                type.add("unknown");
+            }
+            System.out.println("     Now you have " + list.size() + " tasks in the list.");
+            System.out.println("    ____________________________________________________________");
```
joshualeeky: I think cause you used this line a few times, it is possible to declare it as a constant and just call it when you want to print it, it will make the code neater.

src/main/java/Event.java
```diff
@@ -0,0 +1,19 @@
...
+        this.at = at;
+    }
+
+    @Override
+    public String toString() {
+        return "[E]" + super.toString() + " (at: " + at + ")";
+    }
+
+    @Override
+    public String doneToString() {
```
joshualeeky: Maybe you could change this function's name? It could help indicate what the function does.

src/main/java/Duke.java
```diff
@@ -1,10 +1,180 @@
...
+            done.add(0);
+            String description;
+            String time;
+            if (line.startsWith("todo")) {
+                type.add("T");
+                description = line.substring(5);
+                Todo todo = new Todo(description);
+                list.add(description);
+                date.add("");
+                System.out.println("       " + todo);
```
asdzxc270697: I feel that it'll be better to create a method for each commands.

src/main/java/Duke.java
```diff
@@ -1,10 +1,180 @@
...
+import java.util.Scanner;
+
 public class Duke {
+    private static ArrayList<String> list = new ArrayList<>();
+    private static ArrayList<Integer> done = new ArrayList<>();
+    private static ArrayList<String> type = new ArrayList<>();
+    private static ArrayList<String> date = new ArrayList<>();
+
+    private static void greeting() {
+        System.out.println("    ____________________________________________________________");
```
asdzxc270697: Since this is used multiple times in the code, i feel is better to create a constant DIVIDER or something.

asdzxc270697 (commented):
Overall looks good to me!

src/main/java/Duke.java
```diff
@@ -6,16 +6,18 @@
 public class Duke {
     private static ArrayList<String> list = new ArrayList<>();
     private static ArrayList<Integer> done = new ArrayList<>();
+    private static ArrayList<String> type = new ArrayList<>();
+    private static ArrayList<String> date = new ArrayList<>();
 
-    public static void greeting() {
+    private static void greeting() {
```
ruyian: use present tense like greet() instead of greeting to follow better code quality.

src/main/java/Duke.java
```diff
@@ -73,9 +96,39 @@ public static void addList() {
...
+                int seperate = line.indexOf("/by");
+                description = line.substring(9, seperate - 1);
+                time = line.substring(seperate + 4);
+                Deadline deadline = new Deadline(description, time);
+                list.add(description);
+                date.add(time);
+                System.out.println("       " + deadline);
+            } else if (line.startsWith("event")) {
+                type.add("E");
+                int seperate = line.indexOf("/at");
```
ruyian, line 123: Be careful in the indexOf("/at"). Sometimes the input might not be /at but just starting with /

src/main/java/Duke.java
```diff
@@ -84,6 +137,32 @@ public static void addList() {
...
+            System.out.println("       " + todo.doneToString());
+            break;
+        case "D":
+            Deadline deadline = new Deadline(description, time);
+            System.out.println("       " + deadline.doneToString());
+            break;
+        case "E":
+            Event event = new Event(description, time);
+            System.out.println("       " + event.doneToString());
+            break;
```
ruyian: A similar switch statement is found in the main() method. Good to abstract and avoid redundant coding.

ruyian (commented):
Great job, Siyuan!

src/main/java/Duke.java
```diff
@@ -1,10 +1,320 @@
...
+        }
+    }
+
+    private static void createFileOrFolder(String pathName, String fileName) {
+        try {
+            Path path = Paths.get(pathName);
+            Files.createDirectory(path);
+            Path file = Paths.get(pathName + fileName);
+            Files.createFile(file);
+        } catch (FileAlreadyExistsException ignored) {
```
kaiwen98: Suggest to add a comment with the catch block.

src/main/java/Duke.java
```diff
@@ -1,10 +1,320 @@
...
+        } catch (IOException e) {
+            System.out.println("Something went wrong: " + e.getMessage());
+        }
+    }
+
+    private static void appendToFile(String filePath, String textToAppend) throws IOException {
+        FileWriter fw = new FileWriter(filePath, true);
```
kaiwen98: Suggest to replace variable name with a fully spelled one.

src/main/java/Task.java
```diff
@@ -0,0 +1,18 @@
...
+    protected String description;
+
+    public Task(String description) {
+        this.description = description;
+    }
+
+    public String toString(boolean isDone) {
+        String doneSymbol;
+        if (isDone) {
+            doneSymbol = "X";
```
kaiwen98, line 12: Suggest to change "doneSymbol" to be the class attribute, since you may want to persist the state of the variable after an instance of the class is created.

src/main/java/Duke.java
```diff
@@ -1,10 +1,320 @@
...
+import java.util.Scanner;
+import java.util.Set;
+import java.io.FileWriter;
+import java.io.IOException;
+
 public class Duke {
+    private static ArrayList<String> descriptions = new ArrayList<>();
+    private static ArrayList<Boolean> dones = new ArrayList<>();
+    private static ArrayList<String> types = new ArrayList<>();
+    private static ArrayList<String> dates = new ArrayList<>();
```
kaiwen98: Suggest to distribute the logic here to the class definitions. This approach conflicts with OOP.

src/main/java/Duke.java
```diff
@@ -1,10 +1,320 @@
...
+        System.out.println("     Now you have " + descriptions.size() + " tasks in the list.");
+        System.out.println("    ____________________________________________________________");
+    }
+
+    private static void loadTasks() {
+        String pathName = "./data/";
+        String fileName = "duke.txt";
+        File file = new File(pathName + fileName);
+        try {
+            Scanner sc = new Scanner(file);
```
kaiwen98: Suggest to rename the "sc" variable to a fully-spelled one.

src/main/java/Duke.java
```diff
@@ -1,10 +1,320 @@
+// import libraries here
+
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.nio.file.*;
```
kaiwen98: Replace asterisks with an explicit list of imports.

src/main/java/Duke.java
```diff
@@ -1,10 +1,320 @@
...
+            }
+            if (line.startsWith("done")) {
+                int number = Integer.parseInt(line.substring(line.length() - 1));
+                markAsDone(number);
+                saveTasks();
+                line = in.nextLine();
+                continue;
+            }
+            if (line.startsWith("delete")) {
+                int number = Integer.parseInt(line.substring(line.length() - 1));
```
kaiwen98: Suggest to replace variable "number" with a clearer and more verbose name.

PR #157, opened 2021-08-30
[Rakesh] iP

Rakesh12000:


src/main/java/Duke.java
```diff
@@ -1,10 +1,86 @@
...
+
 public class Duke {
     public static void main(String[] args) {
-        String logo = " ____        _        \n"
-                + "|  _ \\ _   _| | _____ \n"
-                + "| | | | | | | |/ / _ \\\n"
-                + "| |_| | |_| |   <  __/\n"
-                + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
+        String logo = "_________________________________________________\n";
```
uosjapuelks: I think for constants like these they can be renamed to be LOGO (all caps) as it seems to be final and you won't be modifying this value.

src/main/java/Duke.java
```diff
@@ -1,10 +1,86 @@
...
+
+        System.out.println(logo + "Hello! I'm Duke\n" + "Whatchu want\n" + logo);
+
+        Scanner in = new Scanner(System.in);
+
+        Task[] Task = new Task[100];
+        String word;
```
uosjapuelks: maybe you should be more specific with the naming of the variables because it affects the readability

uosjapuelks: Changing it to userInput / userCommand may be a better idea to make it easier to read

src/main/java/Duke.java
```diff
@@ -1,10 +1,86 @@
...
+        String date = "";
+        int Count = 0;
```
uosjapuelks: Probably shouldn't start with capital letter, so should change to 'count'

src/main/java/Duke.java
```diff
@@ -1,10 +1,86 @@
...
+
+        do {
+            word = in.nextLine();
+            String[] words = word.split(" "); // stores string of multiple words in a string Array
```
uosjapuelks: word technically has words in it, so again, if you are sticking to using word, then this array should probably be renamed to seperateWords

src/main/java/Duke.java
```diff
@@ -1,10 +1,86 @@
...
+    public static boolean isTodo(String word) {
+        return word.equals("Todo") || word.equals("todo");
+    }
+
+    public static boolean isDeadline(String word) {
+        return word.equals("Deadline") || word.equals("deadline");
+    }
+
+    public static boolean isEvent(String word) {
+        return word.equals("Event") || word.equals("event");
```
uosjapuelks: you could lower the caps of letter of the string inputs to your boolean checking so you need not use '||'

uosjapuelks (commented):
Overall pretty good, Nice Refractoring of the boolean methods to check the commands.
Areas of improvements are mostly naming, the standard for a final CONSTANT.

src/main/java/Duke.java
Uxinnn: I agree as coding standards require variables to be in camelcase

src/main/java/Duke.java
```diff
@@ -1,10 +1,86 @@
+import java.util.Arrays;
+import java.util.Scanner;
```
Uxinnn: Good to import only specific classes needed instead of using *

src/main/java/Duke.java
Uxinnn: Should not be use the name logo since it's not a logo but just a line. Can change to something like "HORIZONTAL_LINE" instead

src/main/java/Duke.java
```diff
@@ -1,10 +1,86 @@
...
+                }
+                else if (isDeadline(words[0])) { //Deadline task
+                    date = word.substring(word.lastIndexOf("/") + 1);
+                    Task[Count] = new Deadline(word, date);
+                    Count++;
+                }
+
+                else if (isEvent(words[0])) { //Event task
+                    date = word.substring(word.lastIndexOf("/") + 1);
+                    Task[Count] = new Event(word, date);
```
Uxinnn: Date is not a good variable name since this variable when used in the Event class can contain location data too.

Uxinnn (commented):
1st PR review done

src/main/java/Duke.java
```diff
@@ -1,10 +1,282 @@
...
+            }
+
+        } while (!isBye(userInput)); //Exit
+
+        System.out.println(LINE + "Bye. Hope to see you again soon!\n" + LINE);
+
     }
+
+    public static boolean isBye(String word) {
+        return word.equalsIgnoreCase("bye");
```
nishanthelango: Consider extracting magic literals such as "bye" into constants.

src/main/java/DukeException.java
```diff
@@ -0,0 +1,111 @@
...
+    protected boolean isNotDone;
+    protected boolean isDoneNoNumber;
+    protected boolean isTodoEmpty;
+    protected boolean isDeadlineEmpty;
+    protected boolean isEventEmpty;
+    protected boolean isDeadlineNoDate;
+    protected boolean isEventNoDate;
+    protected boolean isNotDelete;
+    protected boolean isDeleteNoNumber;
+    protected String Description;
```
nishanthelango, line 17: Consider using camel case for variable names as per coding standards. e.g "Description" can be renamed to "description".

src/main/java/Duke.java
```diff
@@ -1,10 +1,282 @@
...
+    private static boolean hasData = false;
+    public static void main(String[] args) throws FileNotFoundException {
+        String LINE  = "_______________________________________________________________\n";
+
+
+
+        Scanner in = new Scanner(System.in);
+
+        //Task[] Task = new Task[100];
+        ArrayList<Task> Task = new ArrayList<>();
```
nishanthelango: Consider using the plural form for naming collections of objects. e.g "Task" can be renamed to "tasks".

src/main/java/Duke.java
```diff
@@ -1,10 +1,282 @@
...
+                }
+
+                switch(isInvalid) { //handles errors
+                    case 0: System.out.println(LINE + "Ok! I've added this " + firstWord.toLowerCase() + ":\n" +
+                            Task.get(count - 1) + "\n" + "Now you have " + count + (count == 1 ? " task":" tasks") + " in the list.\n" +
+                            "Type 'list' to save / view your tasks!\n" + LINE);
+                            break;
+                    case 1: System.out.println(LINE + invalid + LINE);
+                            break;
+                }
```
nishanthelango: Avoid indenting case clauses. You can configure the switch case format in the IDE.

src/main/java/Duke.java
```diff
@@ -1,10 +1,282 @@
...
-    public static void main(String[] args) {
-        String logo = " ____        _        \n"
-                + "|  _ \\ _   _| | _____ \n"
-                + "| | | | | | | |/ / _ \\\n"
-                + "| |_| | |_| |   <  __/\n"
-                + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
+    private static final ArrayList<String> Data = new ArrayList<>();
+    private static boolean hasData = false;
+    public static void main(String[] args) throws FileNotFoundException {
```
nishanthelango: Avoid long methods (>30 lines) if possible. You can extract some of the logic into methods for better readability.

src/main/java/Duke.java
```diff
@@ -1,10 +1,282 @@
...
+
+        //Task[] Task = new Task[100];
+        ArrayList<Task> Task = new ArrayList<>();
+
+        String userInput;
+        String date = "";
+
+        int count = 0;
+
+        String fileName = "C:\\data\\duke.txt"; //handles data file creation
```
nishanthelango: Consider writing comments above the code, relative to the position of the code, as per coding standards.

src/main/java/Duke.java
```diff
@@ -1,10 +1,282 @@
...
+        scanFileContents(fileName); //scans data
+
+        System.out.println(LINE + "Hello! I'm Duke\n" + "Whatchu want\n" + LINE);
+
+        if (hasData) {
+            for (int i = 0; i < Data.size(); i++) { //transfers data into task array
+                if (Data.get(i).contains("[T]")) {
+                    String task = "todo " + Data.get(i).substring(10);
+                    Task.add(count, new Todo(task));
+                    count++;
```
nishanthelango: Try to avoid arrowhead code and use at most 3 levels of indentation if possible.

PR #161, opened 2021-08-30
[Weichen] iP

jiangweichen835:


src/main/java/Duke.java
```diff
@@ -1,10 +1,54 @@
...
+            System.out.println(LINE + "Bye. Hope to see you again soon!\n");
+        }
+        in.close();
+    }
+
+    public static void MarkasDone(Task[] task, int number) {
+        task[number-1].setDone();
+        System.out.println(LINE + "Nice! I've marked this task as done:\n");
+        System.out.println( task[number-1].getStatusIcon() + " " + task[number-1].getDescription());
+    }
```
ngoivanessa: methods should be placed before main

src/main/java/Duke.java
```diff
@@ -1,10 +1,54 @@
...
+        System.out.println("Hello! I'm Duke\n" + "What can I do for you?\n" + LINE);
+        inputTask();
     }
-}
+
+    public static void inputTask() {
+        Scanner in = new Scanner(System.in);
+        String inputTask = in.nextLine();
+        Task[] tasks = new Task[100];
+        int count = 0;
```
ngoivanessa: naming should be more descriptive, maybe taskCount would be better instead of count

src/main/java/Duke.java
```diff
@@ -1,10 +1,54 @@
...
+                System.out.println(LINE + "added: " + inputTask);
+                tasks[count] = new Task(inputTask);
+                count++;
+            }
+            inputTask = in.nextLine();
+        }
+        if (inputTask.equals("bye")) {
+            System.out.println(LINE + "Bye. Hope to see you again soon!\n");
+        }
+        in.close();
```
ngoivanessa: guard close should be on top, before the while loop

src/main/java/Duke.java
```diff
@@ -1,10 +1,54 @@
...
+    }
+
+    public static void MarkasDone(Task[] task, int number) {
```
ngoivanessa: the large tasks array should not be passed around between methods, since it's in the same class can directly refer to it, don't need to put it as an input parameter. 

naming is also confusing, task and tasks are very similar

ngoivanessa (commented):
overall well written code! just a few minor changes to make

src/main/java/Duke.java
```diff
@@ -1,10 +1,54 @@
...
+        Scanner in = new Scanner(System.in);
+        String inputTask = in.nextLine();
+        Task[] tasks = new Task[100];
+        int count = 0;
+
+        while(!inputTask.equals("bye")) {
+            if (inputTask.equals("list")) {
+                System.out.println(LINE + "Here are the tasks in your list:\n");
+                for (int i = 0; i < count; i++) {
+                    System.out.println((i + 1) + ". " + tasks[i].getStatusIcon() + " " + tasks[i].getDescription());
```
Kair0s3: I feel this line can be changed to be line wrapped instead as it comes really close to the line length limit.

src/main/java/Duke.java
```diff
@@ -1,10 +1,54 @@
...
+                }
+            } else if (inputTask.contains("done")) {
+                int sep = inputTask.indexOf(" ");
+                int number = Integer.parseInt(inputTask.substring(5));
+                MarkasDone(tasks, number);
+            }
+            else {
```
Kair0s3: I feel like it would be nicer to place this line on line 35 instead as this would adhere more towards the coding standards, specifically under layouts section.

src/main/java/Duke.java
```diff
@@ -1,10 +1,54 @@
...
+        if (inputTask.equals("bye")) {
+            System.out.println(LINE + "Bye. Hope to see you again soon!\n");
+        }
+        in.close();
+    }
+
+    public static void MarkasDone(Task[] task, int number) {
+        task[number-1].setDone();
+        System.out.println(LINE + "Nice! I've marked this task as done:\n");
+        System.out.println( task[number-1].getStatusIcon() + " " + task[number-1].getDescription());
```
Kair0s3: Should the space after the opening parenthesis be removed?

src/main/java/Task.java
```diff
@@ -0,0 +1,22 @@
...
+    protected String description;
+    protected boolean isDone;
+
+    public Task(String description) {
+        this.description = description;
+        this.isDone = false;
+    }
+
+    public String getStatusIcon() {
+        return (isDone ? "[X]" : "[ ]"); // mark done task with X
```
Kair0s3: I feel that the comments should be placed on a seperate line relative to the code

src/main/java/Duke.java
Kair0s3: Would it be better to change the function name to be in camelCase?

Kair0s3 (commented):
CS2113T PR 1

src/main/java/Duke.java
peng-217: I think no space is need between the left bracket and task[number-1].

src/main/java/Duke.java
```diff
@@ -1,10 +1,54 @@
...
+        int count = 0;
+
+        while(!inputTask.equals("bye")) {
+            if (inputTask.equals("list")) {
+                System.out.println(LINE + "Here are the tasks in your list:\n");
+                for (int i = 0; i < count; i++) {
+                    System.out.println((i + 1) + ". " + tasks[i].getStatusIcon() + " " + tasks[i].getDescription());
+                }
+            } else if (inputTask.contains("done")) {
+                int sep = inputTask.indexOf(" ");
```
peng-217: For integer "sep", it would be better to give it a clearer name so that readers can directly know what it is used for.

src/main/java/Duke.java
```diff
@@ -1,10 +1,54 @@
+import java.util.Scanner;
+
 public class Duke {
+
```
peng-217, line 9: It would be better to write some comments to illustrate what are the functions of the methods

src/main/java/Duke.java
```diff
@@ -1,10 +1,54 @@
...
+        while(!inputTask.equals("bye")) {
+            if (inputTask.equals("list")) {
+                System.out.println(LINE + "Here are the tasks in your list:\n");
+                for (int i = 0; i < count; i++) {
+                    System.out.println((i + 1) + ". " + tasks[i].getStatusIcon() + " " + tasks[i].getDescription());
+                }
+            } else if (inputTask.contains("done")) {
+                int sep = inputTask.indexOf(" ");
+                int number = Integer.parseInt(inputTask.substring(5));
+                MarkasDone(tasks, number);
```
peng-217: I think the name of method MarkasDone should be markAsDone?

peng-217 (commented):
The code is generally good, it could be better if you can add some comment of each method and pay attention to the naming style.

src/main/java/Duke.java
```diff
@@ -1,10 +1,171 @@
...
+
+    private static void addDeadline(String inputTask) {
+        try {
+            String description = inputTask.substring(9);
+            int index = description.indexOf("/by");
+            String deadlineDescription = description.substring(index + 4);
+            String taskDescription = description.substring(0, index);
+            tasks.add(new Deadline(taskDescription, deadlineDescription));
+
+            System.out.println("Got it. I've added this task:\n" + "[D][]" + taskDescription + " (by: " + deadlineDescription + ")\n" + "Now you have " + tasks.size() + " tasks in this list.");
```
OngDeZhi: Line length should be no longer than 120 chars, consider using another print statement. Similarly on line 148.

src/main/java/Duke.java
```diff
@@ -1,10 +1,171 @@
...
+            } catch (DukeException e) {
+                System.out.println("☹ OOPS!!! I'm sorry, but I don't know what that means :-( ");
+                System.out.println(LINE);
+            }
+            inputTask = in.nextLine();
+        }
+        if (inputTask.equals("bye")) {
+            try {
+                writeToFile();
+            } catch (IOException e) {}
```
OngDeZhi: Avoid having an empty catch block. If you really want to leave empty, then leave some comment as to why it is left empty.

src/main/java/Duke.java
```diff
@@ -1,10 +1,171 @@
...
+                    task.setDone();
+                }
+            }
+            s.close();
+        } catch (FileNotFoundException e) {
+            System.out.println("Unable to find file\n");
+        }
+    }
+
+    public static void inputTask() {
```
OngDeZhi: The method is pretty long, may want to create a new method for some of the statement.

src/main/java/Duke.java
```diff
@@ -1,10 +1,171 @@
...
+
+    private static void writeToFile() throws IOException {
+        FileWriter fw = new FileWriter(file);
+        for (int i = 0; i < tasks.size(); i++) {
+            fw.write(tasks.get(i).toString() + System.lineSeparator());
+        }
+        fw.close();
+    }
+
+    private static void saveToFile() throws IOException{
```
OngDeZhi: Is this method used, I don't see it being used anywhere and I'm also not quite sure if it's complete (seems to be half done). You may also want to comment on the difference between writeToFile and saveToFile, looks to be the same purpose to me.

src/main/java/Duke.java
```diff
@@ -1,10 +1,171 @@
...
+                } else if (inputTask.contains("deadline")) {
+                    addDeadline(inputTask);
+                } else if (inputTask.contains("event")) {
+                    addEvent(inputTask);
+                } else if (inputTask.contains("delete")) {
+                    deleteTask(inputTask);
+                }
+                else {
+                    throw new DukeException();
+                }
```
OngDeZhi: Can apply SLAP to this set of statements.
Also, you may want to use a switch statement instead, it will help make it look cleaner for your readers.

PR #163, opened 2021-08-30
[Hu Jialun] iP

SuibianP:


src/main/java/Duke.java
```diff
@@ -1,10 +1,185 @@
...
+        private String title = "";
+        private boolean doneStatus = false;
+
+        protected String getType() {
+            return "task";
+        }
+
+        private Task() {
+            // prevent uninitialised task
+        }
```
xuanlc113: i'm not really sure what private constructor is for

src/main/java/Duke.java
```diff
@@ -1,10 +1,185 @@
...
+        String assist = "What can I do for you?";
+        System.out.println(greeting);
+        System.out.println(assist);
+    }
+
+    private enum InputHandleStatus {
+        OK, END, ERROR
+    }
+
+    static class Task {
```
xuanlc113: Would it be better to put this in another file

src/main/java/Duke.java
```diff
@@ -1,10 +1,185 @@
...
+            case "todo":
+                description = String.join(" ", Arrays.copyOfRange(args, 1, args.length));
+                newTask = new Todo(description);
+                break;
+        }
+        return newTask;
+    }
+
+    private static InputHandleStatus handleOneInputLine(String line) {
+        String[] splitted = line.split("\\s+");
```
xuanlc113: Should this variable be plural, since its an array?

src/main/java/Duke.java
```diff
@@ -1,10 +1,185 @@
...
+    private static void initialise() {
+        String greeting = "Hello! I'm Duke";
+        String assist = "What can I do for you?";
+        System.out.println(greeting);
+        System.out.println(assist);
+    }
+
+    private enum InputHandleStatus {
+        OK, END, ERROR
+    }
```
Alvinlj00: I like the use of enumeration to handle potential errors with clear constants

src/main/java/Duke.java
```diff
@@ -1,10 +1,185 @@
...
-                + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
+        initialise();
+        Scanner sc = new Scanner(System.in);
+        while (sc.hasNextLine()) {
+            switch (handleOneInputLine(sc.nextLine())) {
+            case END:
+                finalise();
+                return;
+            }
```
Alvinlj00: Could a default case be added as well?

src/main/java/Duke.java
```diff
@@ -1,10 +1,185 @@
...
+                int atPosition = Arrays.asList(args).indexOf("/at");
+                description = String.join(" ", Arrays.copyOfRange(args, 1, atPosition));
+                String at = String.join(" ", Arrays.copyOfRange(args, atPosition + 1, args.length));
+                newTask = new Event(description, at);
+                break;
+            case "todo":
+                description = String.join(" ", Arrays.copyOfRange(args, 1, args.length));
+                newTask = new Todo(description);
+                break;
+        }
```
Alvinlj00: Could a default case be added as well?

src/main/java/Duke.java
```diff
@@ -1,10 +1,185 @@
...
+                target.setDoneStatus(true);
+                System.out.println("Nice! I've marked this task as done:");
+                System.out.println(target);
+                return InputHandleStatus.OK;
+            default:
+                Task newTask = parseTask(splitted);
+                taskList.add(newTask);
+                System.out.println("added: " + newTask);
+                return InputHandleStatus.OK;
+        }
```
Alvinlj00: I like the use of switch cases to filter the type of task here

src/main/java/Duke.java
```diff
@@ -1,10 +1,185 @@
...
+            return super.toString() + " (at: " + getTo() + ")";
+        }
+
+        @Override
+        protected String getType() {
+            return "event";
+        }
+    }
+
+    private static void finalise() {
```
Alvinlj00: Could 'terminate' be a better choice than 'finalise'?

Alvinlj00 (commented):
Overall, there were not many coding standard violations, but remember to add default cases to switch statements.

src/main/java/Duke.java
SuibianP: Fixed in 4b1cb6f

src/main/java/Duke.java
SuibianP: `splitteds` sounds weird :thinking: Any suggestions?

src/main/java/Duke.java
SuibianP: I also weighed it and came across [this SO question](https://softwareengineering.stackexchange.com/questions/163004/what-is-the-opposite-of-initialize-or-init). Did not choose `terminate` since it sounds like it is the last thing while it might no longer be in future (e.g. might have some other cleanup after saying goodbye).

src/main/java/Duke.java
SuibianP: If it is not a valid task type, returning `null` seems sensible already :thinking:

src/main/java/Duke.java
xuanlc113: maybe something like commandParams?

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,104 @@
+package duke;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Scanner;
+import java.io.*;
```
Tejasvas123: Avoid wildcard import statements for packages. Explicitly mention all the classes being used.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,104 @@
...
+
+import duke.task.*;
+import duke.command.TaskParser;
+
+public class Duke {
+
+    static ArrayList<Task> taskList = new ArrayList<>();
+    static final String FILENAME = "tasklist";
```
Tejasvas123: Multiple words like `FILENAME` can be written as `FILE_NAME`

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,104 @@
...
+
+    private static void saveTaskListToFile() {
+	try {
+	    FileOutputStream fileos = new FileOutputStream(FILENAME);
+	    ObjectOutputStream objos = new ObjectOutputStream(fileos);
```
Tejasvas123: Save your file contents in human readable format (instead of directly saving and loading from objectstream).

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,104 @@
...
+        }
+	return false;
+    }
+
+    private static void initialise() {
+	readTaskListFromFile();
+        String greeting = "Hello! I'm Duke";
+        String assist = "What can I do for you?";
+        System.out.println(greeting);
+        System.out.println(assist);
```
Tejasvas123: Inconsistent level of indentation here

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    public static void main(String[] args) {
+        initialise();
+        Scanner sc = new Scanner(System.in);
+        while (sc.hasNextLine()) {
+	    try {
+                if (handleOneInputLine(sc.nextLine())) {
+                    finalise();
+		    return;
```
Tejasvas123: Inconsistent level of indentation

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,104 @@
...
+	} catch (IOException | ClassNotFoundException e) {
+	    System.out.println("Error reading from file.");
+	    System.out.println(e.getMessage());
+	}
+    }
+
+    private static boolean handleOneInputLine(String line) {
+        String[] splitted = line.split("\\s+");
+        switch (splitted[0]) {
+            case "bye":
```
Tejasvas123: String constants like `"bye"`, `"list"` should be added as named constants, for example: `BYE_COMMAND = "bye"`

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,104 @@
...
+	    FileInputStream fileis = new FileInputStream(FILENAME);
+	    ObjectInputStream objis = new ObjectInputStream(fileis);
+	    taskList = (ArrayList<Task>) objis.readObject();
+	} catch (IOException | ClassNotFoundException e) {
+	    System.out.println("Error reading from file.");
+	    System.out.println(e.getMessage());
+	}
+    }
+
+    private static boolean handleOneInputLine(String line) {
```
Tejasvas123, line 10: Perhaps a better approach would be handle the parsing of inputs in a separate class like `Parser` which returns the input in the desired format to execute the commands

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,104 @@
...
+    static ArrayList<Task> taskList = new ArrayList<>();
+    static final String FILENAME = "tasklist";
+
+    private static void saveTaskListToFile() {
+	try {
+	    FileOutputStream fileos = new FileOutputStream(FILENAME);
+	    ObjectOutputStream objos = new ObjectOutputStream(fileos);
+	    objos.writeObject(taskList);
+	} catch (IOException e) {
+	    System.out.println("Error saving to file.");
```
Tejasvas123: Error strings are also considered as constants and should be added as named constants

src/main/java/duke/Duke.java
Tejasvas123: There should be no indentation for the `case blocks`:
```
switch(splitted[0]) {
case "bye": ....

```

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,104 @@
...
+                System.out.println("Nice! I've marked this task as done:");
+                System.out.println(target);
+		break;
+	    }
+	    case "delete": {
+                Task target = taskList.remove(Integer.parseInt(splitted[1]) - 1);
+		saveTaskListToFile();
+		System.out.println("Noted. I've removed this task:");
+		System.out.println(target);
+		break;
```
Tejasvas123: Inconsistent level of indentation

src/main/java/duke/command/TaskParser.java
```diff
@@ -0,0 +1,43 @@
+package duke.command;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.ArrayList;
+import duke.task.*;
```
Tejasvas123: Avoid wildcard import statements

Tejasvas123 (commented):
Great job Jialun!

PR #164, opened 2021-08-30
[Koh Qianqi] iP

qqkoh:


src/main/java/Duke.java
```diff
@@ -43,10 +52,9 @@ private static String getUserInput() {
         return input;
     }
 
-    // first element is command type, second is arguments string
     private static String[] splitCommandWordsAndArgs(String rawUserInput) {
```
tryyang2001: I think you can consider adding documentation for each method implemented.

src/main/java/Duke.java
```diff
@@ -1,24 +1,33 @@
 import java.util.Scanner;
+import java.util.regex.Pattern;
 
 public class Duke {
     private static final Scanner scan = new Scanner(System.in);
-    private static final String KEYWORD_BY = "/by";
-    private static final String KEYWORD_AT = "/at";
+    private static final String LINE_PREFIX = "\t";
```
tryyang2001: I would consider using this directly without declaring it as a constant variable?

src/main/java/Duke.java
```diff
@@ -43,10 +52,9 @@ private static String getUserInput() {
         return input;
     }
 
-    // first element is command type, second is arguments string
     private static String[] splitCommandWordsAndArgs(String rawUserInput) {
         final String[] split = rawUserInput.trim().split("\\s+", 2);
-        return split.length == 2 ? split : new String[]{split[0], ""}; // else case: no parameters
+        return split.length == 2 ? split : new String[]{split[0], ""};
```
tryyang2001: I think using ternary operation might be more confusing as compared to if-else statement with return value "true" or "false" under this method

src/main/java/Duke.java
```diff
@@ -1,24 +1,33 @@
...
     private static final String MESSAGE_INVALID_COMMAND_FORMAT = "Invalid command format.";
     private static final String MESSAGE_EMPTY_TASK_LIST = "There are no tasks in your to-do list. Bet that'll change soon.";
     private static final String MESSAGE_TASKS_FOUND_OVERVIEW = "There are %1$d %2$s in your list";
-    private static final String LINE_PREFIX = "\t";
-    private static final String LS = System.lineSeparator() + LINE_PREFIX;
+    private static final String MESSAGE_BYE = "bye. see you never again.";
+    private static final String MESSAGE_CORRECT_FORMAT_ADD_TODO = "add todo format:" + INDENTED_NEW_LINE + "todo wash the dishes";
+    private static final String MESSAGE_CORRECT_FORMAT_ADD_EVENT = "add event format: " + INDENTED_NEW_LINE + "event (description) /at (time)";
+    private static final String MESSAGE_CORRECT_FORMAT_ADD_DEADLINE = "add deadline format: " + INDENTED_NEW_LINE + "deadline (description) /by (time)";
+    private static final String MESSAGE_CORRECT_FORMAT_MARK_AS_DONE = "mark task as completed format: " + INDENTED_NEW_LINE + "done (task number)";
```
tryyang2001: Good effort on declaring all these constant values!

src/main/java/Duke.java
```diff
@@ -135,44 +159,46 @@ private static void addTask(Task t) {
     private static String executeAddTodo(String description) {
         addTask(new Todo(description));
         todosCount++;
-        return "added: " + tasks[tasksCount - 1].toString() + LS + String.format(MESSAGE_TASKS_FOUND_OVERVIEW, todosCount, COMMAND_TODO_WORD);
+        return MESSAGE_ADDED + tasks[tasksCount - 1].toString() + NEW_LINE
+                + String.format(MESSAGE_TASKS_FOUND_OVERVIEW, todosCount, COMMAND_TODO_WORD);
     }
 
     private static String executeAddEvent(String commandArgs) {
         String[] eventAndDate = decodeTask(COMMAND_EVENT_WORD, commandArgs);
```
tryyang2001: I think according to the coding standard we learnt, the name for the string array should be in plural form, so I guess this should be rename as eventAndDates :)

src/main/java/Duke.java
shivin9: I would prefer the current way as the LINE_PREFIX variable can be something other than a `\t` also

src/main/java/Duke.java
```diff
@@ -1,10 +1,335 @@
...
+            switch (taskType) {
+            case 'D':
+                String by = split[3].trim();
+                addTask(new Deadline(description, isDone, by));
+                return;
+            case 'E':
+                String at = split[3].trim();
+                addTask(new Event(description, isDone, at));
+                return;
+            default:
```
shivin9: Nice use of case statements

src/main/java/Duke.java
```diff
@@ -1,10 +1,335 @@
...
+        case COMMAND_HELP_WORD:
+            return getUserGuide();
+        case COMMAND_DELETE_WORD:
+            return executeDeleteTask(commandArgs);
+        case COMMAND_CLEAR_HISTORY:
+            return executeClearHistory();
+        case COMMAND_GET_STATS:
+            return executeGetStats();
+        case COMMAND_EXIT_WORD:
+            exitProgram();
```
shivin9: Indentation?

src/main/java/Duke.java
```diff
@@ -1,10 +1,335 @@
...
+        try {
+            Files.createDirectories(Paths.get(PARENT_PATH));
+            File file = new File(FILEPATH);
+            if (!file.exists()) {
+                file.createNewFile();
+            }
+            FileWriter fw = new FileWriter(file, false);
+            BufferedWriter bw = new BufferedWriter(fw);
+
+            for (Task task : taskList) {
```
shivin9: Nice use of iterators :)

shivin9 (commented):
Nice code overall.

PR #174, opened 2021-08-31
[Hu Xuefei] iP

Xuefei2001:


src/main/java/Duke.java
```diff
@@ -1,10 +1,115 @@
...
                 + "|  _ \\ _   _| | _____ \n"
                 + "| | | | | | | |/ / _ \\\n"
                 + "| |_| | |_| |   <  __/\n"
                 + "|____/ \\__,_|_|\\_\\___|\n";
         System.out.println("Hello from\n" + logo);
+        System.out.println("____________________________________________________________");
+        System.out.println("Hello! I'm Duke");
+        System.out.println("What can I do for you?");
+        System.out.println("type in your command please");
+        System.out.println("____________________________________________________________");
```
powzx: suggestion:
you can refactor the lines to a constant such as 'PROMPT_MESSAGE'

src/main/java/Duke.java
```diff
@@ -1,10 +1,115 @@
...
+    }
+
+    public static void printBye() {
+        System.out.println("____________________________________________________________");
+        System.out.println("Bye. Hope to see you again soon!");
+        System.out.println("____________________________________________________________");
```
powzx: suggestion:
you can extract the horizontal line as a constant 'DIVIDER' if you are going to use it many times in your code.

src/main/java/Duke.java
```diff
@@ -1,10 +1,115 @@
...
+                String description = command.substring(9, slashIndex-1);
+                String by = command.substring(slashIndex+4);
+                tasks[size-1] = new Deadline(description, size, by);
+                printDeadline(tasks[size-1], size);
+            } else if (command.startsWith("event ")) {
+                size++;
+                int slashIndex = command.indexOf("/");
+                String description = command.substring(6, slashIndex-1);
+                String at = command.substring(slashIndex+4);
+                tasks[size - 1] = new Event(description, size, at);
```
powzx: Using switch-case statements may make the code look cleaner and easier to follow

src/main/java/Duke.java
```diff
@@ -1,10 +1,115 @@
...
+                setTaskDone(tasks[i]);
+                break;
+            }
+        }
+    }
+
+    public static void main(String[] args) {
+        printGreeting();
+        int size = 0;
+        Task[] tasks = new Task[100];
```
powzx: You can extract 100 as MAX_NUMBER_OF_TASKS as other readers may not know how you derived the number 100

src/main/java/Duke.java
gavienwz: Seconded

src/main/java/Duke.java
gavienwz: Seconded

src/main/java/Deadline.java
```diff
@@ -0,0 +1,17 @@
+public class Deadline extends Task{
+    protected String by;
```
gavienwz: String can have more descriptive name

src/main/java/Duke.java
```diff
@@ -1,10 +1,115 @@
...
+        System.out.println("____________________________________________________________");
+    }
+
+    public static void printEvent(Task task, int size) {
+        System.out.println("____________________________________________________________");
+        System.out.println("Got it. I've added this task: \n" + "\t" + task.toString());
+        System.out.println("Now you have " + size + " tasks in the list.");
+        System.out.println("____________________________________________________________");
+    }
+
```
gavienwz: Since these print methods all print the same message, it may be better to have a single general method

gavienwz (commented):
Overall code is neat, but naming can be more descriptive for readers

src/main/java/Duke.java
ZhangErliCarl: A single general method to print all these message might be better for easier reading.

src/main/java/Duke.java
ZhangErliCarl: It is better not to include the "magic number" in the code, eg. 100 here. It might be better to extract it as a variable and then use it in the codeing

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,513 @@
...
+            throw new NoSpaceException();
+        } else if ((command.length() == slashIndex + 3)) {
+            throw new NoTimeException();
+        } else if ((command.length() == (slashIndex + 4)) && (command.charAt(slashIndex + 3) == ' ')) {
+            throw new NoTimeException();
+        } else if (command.charAt(slashIndex + 3) != ' ') {
+            throw new NoSpaceException();
+        } else if (command.contains("/at")) {
+            throw new WrongTimeTypeException();
+        }
```
Tejasvas123, line 128: may be can convert this chunk of code to the correspond switch case block for ease of reading

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,513 @@
...
+        for (String taskLine : taskLines) {
+            try {
+                appendToFile(taskLine + System.lineSeparator());
+            } catch (IOException e) {
+                System.out.println("Something went wrong: " + e.getMessage());
+            }
+        }
+    }
+
+    public static void main(String[] args){
```
Tejasvas123, line 465: Avoid methods with longer than 30 lines. This method could perhaps be modularised and broken down into simpler constructs.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,513 @@
...
+        printGreeting();
+        initializeHardDiskSave();
+        ArrayList<Task> tasks = new ArrayList<>();
+        ArrayList<String> taskLines = new ArrayList<>();
+        try {
+            initializeTaskLineList(taskLines);
+            initializeList(tasks);
+        } catch (FileNotFoundException e) {
+            System.out.println("File not found, pls try again");
```
Tejasvas123, line 474: Error messages are also considered as String constants so pleased add them as named constants

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,513 @@
...
+        }
+
+        Scanner in = new Scanner(System.in);
+        while (true) {
+            String command = in.nextLine();
+            if (command.equals("bye")) {
```
Tejasvas123, line 480: String literals like `"bye", "echo"` are considered as constants so perhaps you can store them as constants for better readabililty

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,513 @@
...
+
+    public static String convertTodotoLine(ToDo task) {
+        if (task.isDone()) {
+            return ("T | √ | " + task.getDescription());
+        } else {
+            return ("T | X | " + task.getDescription());
+        }
+    }
+
+    public static String convertDeadlinetoLine(Deadline task) {
```
Tejasvas123, line 405: Perhaps a better approach (more OOP oriented) would be to add these functions to the respective classes of Events, ToDo etc as they provide functionalities related to those classes

Tejasvas123: Another design approach that you can consider for your Duke would be to have separate classes like 'DataManager' that manages loading and storing of data into the files and `TaskManager` for managing the tasks instead of adding everything to the Duke class

Tejasvas123 (commented):
Great job Xuefei! Suggested some improvement for code quality, design and avoiding coding standard violations.

PR #179, opened 2021-09-01
[Wang Chuhan] ip

Vincent6896:


src/main/java/List.java
```diff
@@ -0,0 +1,81 @@
+import java.lang.String;
+
+public class List {
```
yeezao: I wouldn't recommend calling your class `List` - there's already a provided class within Java called `List`, and this could prove very confusing to readers. You might want to consider changing the name of this class to something else.

src/main/java/List.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    public static void addTask(String task) {
+        if (task.indexOf("/") != -1) {
+            list[count] = (count + 1) + ".["
+                + task.toUpperCase().charAt(0)
+                + "][ ] "
+                + task.substring(task.indexOf(" "), task.indexOf("/"))
+                + "(" + task.substring(task.indexOf("/") + 1, task.indexOf("/") + 3)
+                + ": " + task.substring(task.indexOf("/") + 4) + ")";
```
yeezao: This section looks quite confusing - what are the `+1` and `+3` and `+4` doing there - what do they mean? You might want to define these magic variables before using them.

src/main/java/List.java
```diff
@@ -0,0 +1,81 @@
...
+                    + list[count - 1].substring(2)
+                    + "\n    Now you have " + count + " tasks in the list."
+                    + "\n    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~");
+        }
+    }
+
+    public static void doneTask(String task) {
+        task = task.trim();
+        String str = "";
+        if (task != null && !"".equals(task)) {
```
yeezao: You could consider replacing `"".equals(task)` with `task.isEmpty()` - it reads more sensibly.

src/main/java/List.java
```diff
@@ -0,0 +1,81 @@
...
+    public static String[] list = new String[100];
+    public static String[] done = new String[100];
+    public static int count = 0;
+    public static int doneCount = 0;
+
+    public static void printTask() {
+        System.out.println("    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~\n     "
+                + "Here are the tasks in your list:");
+        for (int i = 0; i < count; i++) {
+            System.out.println("     " + list[i]);
```
yeezao: You could consider defining some variable to deal with the huge block of spaces inside your `println`.

src/main/java/List.java
```diff
@@ -0,0 +1,81 @@
+import java.lang.String;
+
+public class List {
+
+    public static String[] list = new String[100];
+    public static String[] done = new String[100];
```
MaifengNg: Would be good to rename done to something like doneList

src/main/java/Duke.java
```diff
@@ -6,5 +10,36 @@ public static void main(String[] args) {
...
                 + "|____/ \\__,_|_|\\_\\___|\n";
         System.out.println("Hello from\n" + logo);
+
+        System.out.println("    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~\n"
+                + "    Hello! I'm Duke\n"
+                + "    What can I do for you?\n"
+                + "    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~");
+        while (true) {
+            Scanner in = new Scanner(System.in);
+            String order;
```
MaifengNg: variable name order does not break the coding standard but it seems very unintuitive to me, a better name would be maybe input or task

src/main/java/Duke.java
```diff
@@ -6,5 +10,36 @@ public static void main(String[] args) {
...
+                continue;
+            }
+            if (order.startsWith("done")) {
+                List.doneTask(order);
+                continue;
+            }
+            List.addTask(order);
+            if (order.equals("list")) {
+                List.printTask();
+            }
```
MaifengNg: Can consider using case inside of multiple ifs

src/main/java/List.java
```diff
@@ -0,0 +1,81 @@
...
+                + "(" + task.substring(task.indexOf("/") + 1, task.indexOf("/") + 3)
+                + ": " + task.substring(task.indexOf("/") + 4) + ")";
+        } else {
+            list[count] = (count + 1) + ".["
+                + task.toUpperCase().charAt(0)
+                + "][ ] "
+                + task.substring(task.indexOf(" "));
+        }
+        count++;
+        if (count == 1) {
```
MaifengNg: can consider putting this whole chunk in a method to reduce repeating statements

MaifengNg: Could also use variables to represent the repeated print statements

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,324 @@
...
+                    } catch (DukeException e) {
+                        System.out.println(SPLIT_LINE);
+                        System.out.println("    ☹ OOPS!!! I'm sorry, but I don't know what that means :-(");
+                        System.out.println(SPLIT_LINE);
+                        continue;
+                    }
+                }
+                System.out.println(SPLIT_LINE);
+                System.out.println("    Got it. I've added this task:");
+                System.out.println("     " + list.get(list.size() - 1));
```
iamabhishek98: Avoid having such nested logic in print statements as the purpose of print statements is only to print and not do anything else. Seperate this out into separate lines

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,324 @@
...
+                        case "E":
+                            String at = tasks[i].substring(tasks[i].indexOf(":") + 2, tasks[i].indexOf(")"));
+                            if (tasks[i].contains("X")) {
+                                Event event = new Event(description.substring(0, description.indexOf("(") - 1), at);
+                                event.description = event.description.replaceFirst(" ", "X");
+                                event.isDone = true;
+                                list.add(event);
+                            } else {
+                                list.add(new Event(description.substring(0, description.indexOf("(") - 1), at));
+                            }
```
iamabhishek98: Consider abstracting out the logic of each switch case statement into separate functions

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,324 @@
...
+                            continue;
+                        case "E":
+                            String at = tasks[i].substring(tasks[i].indexOf(":") + 2, tasks[i].indexOf(")"));
+                            if (tasks[i].contains("X")) {
+                                Event event = new Event(description.substring(0, description.indexOf("(") - 1), at);
+                                event.description = event.description.replaceFirst(" ", "X");
+                                event.isDone = true;
+                                list.add(event);
+                            } else {
+                                list.add(new Event(description.substring(0, description.indexOf("(") - 1), at));
```
iamabhishek98: As mentioned before, do not use nested logic like this as it makes code extremely hard to read.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,324 @@
...
+                    String[] tasks = str.split(",");
+                    for (int i = 0; i < tasks.length; i++) {
+                        String description = tasks[i].substring(8);
+                        if (i == tasks.length - 1) {
+                            description = description.substring(0, description.length() - 1);
+                        }
+                        switch (tasks[i].substring(2, 3)) {
+                        case "T":
+                            Todo todo = new Todo(description);
+                            todo.description = todo.description.replaceFirst(" ", "X");
```
iamabhishek98: Avoid magic strings like "X" - extract them out as constants and name them appropriately using proper naming conventions

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,324 @@
...
+            System.out.println(SPLIT_LINE + "\n    Here are the tasks in your list:");
+            for (int i = 0; i < list.size(); i++) {
+                System.out.println("    " + (i + 1) + "." + list.get(i));
+            }
+        }
+        System.out.println(SPLIT_LINE);
+    }
+
+    public static void printDoneList() {
+        if (doneList.size() == 0) {
```
iamabhishek98: You can also do something like:
if (!doneList.size()) {
}

src/main/java/duke/List.java
```diff
@@ -0,0 +1,16 @@
+package duke;
+import java.lang.String;
+
+public class List {
+    protected String description;
+    protected boolean isDone;
+
+    public List(String description) {
+        this.description = "[ ] " + description;
```
iamabhishek98: As mentioned earlier, avoid magic strings like "[ ] "

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,324 @@
...
+
+    public static void doneTask(String task) {
+        task = task.trim();
+        String str = "";
+        if (task != null && !"".equals(task)) {
+            for (int i = 0; i < task.length(); i++) {
+                if (task.charAt(i) >= 48 && task.charAt(i) <= 57) {
+                    str = str + task.charAt(i);
+                }
+            }
```
iamabhishek98: Having so many nested if statements and loops can be hard to read. Consider abstracting it out into a different function

iamabhishek98 (commented):

PR #180, opened 2021-09-01
[Huang Che Yen] iP

KaiserHuang88:


src/main/java/Duke.java
```diff
@@ -1,10 +1,85 @@
+import java.util.Arrays;
+import java.util.Scanner;
+
 public class Duke {
+
+    private static boolean diff_string_check(String[] arr, String toCheckValue)
+    {
```
richwill28: This seems to be a violation of the [coding standard](https://se-education.org/guides/conventions/java/index.html#layout), more specifically consider using K&R style brackets. I noticed the same issue in several places below too.

src/main/java/Duke.java
```diff
@@ -1,10 +1,85 @@
+import java.util.Arrays;
```
richwill28: Unused import. Consider removing this line, it would make the code cleaner.

src/main/java/Duke.java
```diff
@@ -1,10 +1,85 @@
...
+                System.out.println("Bye. Hope to see you again soon!");
+                break;
+            }
+
+            // List down
+            if (line.equals("list"))
+            {
+                for (int i = 0; i < total_tasks; i += 1)
+                {
+                    System.out.print((i+1) + "." + "[" + schedule[i].getStatusIcon() + "] ");
```
richwill28: `+` is not followed and preceded by whitespace. This is a violation of the [coding standard](https://se-education.org/guides/conventions/java/index.html#layout), more specifically the `White space within a statement` guideline.

richwill28 (commented):
Overall, code looks good. There are only several coding standard violations.

src/main/java/Task.java
```diff
@@ -0,0 +1,26 @@
...
+        this.isDone = false;
+    }
+
+    public Task()
+    {
+        this.description = "empty";
+        this.isDone = false;
+    }
+
+    public void markAsDone()
```
richwill28: Consider writing a `JavaDoc` comment for public methods, as it is meant to be used by others. This seems to be a violation of this [coding standard](https://se-education.org/guides/conventions/java/index.html#comments), more specifically it is advisable to write descriptive header comments for all public classes/methods.

richwill28 (commented):
Code looks good so far. One improvement to be consider is to create subclasses of `Task`. For example, adding `Todo`, `Deadline`, and `Event` classes to improve the feature of the app.

src/main/java/Duke.java
```diff
@@ -1,10 +1,85 @@
+import java.util.Arrays;
+import java.util.Scanner;
+
 public class Duke {
+
+    private static boolean diff_string_check(String[] arr, String toCheckValue)
+    {
+        boolean test = true;
```
aditichadha1310: You might want to name a boolean variable with something that starts with is or has? It will then follow the coding standards.

src/main/java/Duke.java
```diff
@@ -1,10 +1,85 @@
...
+            if (toCheckValue.equals(arr[i]))
+            {
+                test = false;
+                break;
+            }
+            else if (arr[i] == null)
+            {
+                break;
+            }
+        }
```
aditichadha1310: Perhaps this violates the coding standards. You can write the code as follows:
if(){
---
--
} else if(){
---
---
} else{
--
--
}

src/main/java/Task.java
```diff
@@ -1,2 +1,26 @@
...
+    {
+        this.isDone = true;
+    }
+
+    public String getStatusIcon()
+    {
+        return (isDone ? "X" : " "); // mark done task with X
+    }
```
Mist0919: The first "{" of each method should be next to the name of method.
eg. public String getStatusIcon() {
       ......
}

src/main/java/Duke.java
```diff
@@ -1,10 +1,85 @@
+import java.util.Arrays;
+import java.util.Scanner;
+
 public class Duke {
+
+    private static boolean diff_string_check(String[] arr, String toCheckValue)
+    {
+        boolean test = true;
+
+        for (int i = 0; i < arr.length; i++)
```
xseh: Follow Egyptian style brackets.

src/main/java/Duke.java
```diff
@@ -1,10 +1,85 @@
...
+            {
+                for (int i = 0; i < total_tasks; i += 1)
+                {
+                    System.out.print((i+1) + "." + "[" + schedule[i].getStatusIcon() + "] ");
+                    System.out.println(schedule[i].description);
+                }
+            }
+            
+            //Done
+            else if ((line.length() > 4) && line.substring(0, 4).equals("done"))
```
xseh: Avoid complicated expressions. Consider splitting the complicated expressions into steps. or use SLAP to extract the logic into separate functions.

src/main/java/Duke.java
```diff
@@ -1,10 +1,85 @@
...
+            {
+                int number = Character.getNumericValue(line.charAt(5));
```
xseh: Consider a more meaningful name than `number` which can better describe the value. 

Also avoid the use of magic literals. Using a suitable variable name for the constant numbers (which describes the value) would help the understanding of the program.

src/main/java/Duke.java
```diff
@@ -1,10 +1,85 @@
...
+                schedule[number - 1].markAsDone();
+            }
+
+            // Check for duplicates and add accordingly
```
xseh: Consider removing the comment since the code seems self-explanatory.

src/main/java/Duke.java
```diff
@@ -1,10 +1,85 @@
...
+            else if (diff_string_check(task_list, line))
+            {
+                Task t = new Task(line);
```
xseh: Consider using a more suitable name (noun) instead of an alphabet for the variable.

src/main/java/Duke.java
```diff
@@ -1,10 +1,85 @@
...
+                schedule[total_tasks] = t;
+                task_list[total_tasks] = line;
```
xseh: Variable names must be in camelCase.

src/main/java/Duke.java
```diff
@@ -1,10 +1,85 @@
...
+            
+            //Done
+            else if ((line.length() > 4) && line.substring(0, 4).equals("done"))
+            {
+                int number = Character.getNumericValue(line.charAt(5));
+                schedule[number - 1].markAsDone();
+            }
+
+            // Check for duplicates and add accordingly
+            else if (diff_string_check(task_list, line))
```
xseh: Names representing methods must be verbs and written in camelCase.

xseh (commented):
Good try, consider following the coding standard a little more and SLAP more

PR #186, opened 2021-09-01
[Yap Joon Siong] iP

yapjoonsiong:


src/main/java/Tasks.java
```diff
@@ -0,0 +1,6 @@
+public class Tasks {
+    Task list[] = new Task[100];
+    int listLength = 0;
+    int tasksIncomplete = 0;
```
teoziyiivy: incompleteTasks might be a better variable name here.

src/main/java/Duke.java
```diff
@@ -1,10 +1,133 @@
...
+        user.listLength ++;
+        user.tasksIncomplete ++;
+        System.out.println("\tNow you have " + user.listLength + " tasks in the list, " + user.tasksIncomplete + " incomplete tasks");
+    }
+    public static void markDone(String input, Tasks user) {
+        int index = Integer.parseInt(input.substring(5)) - 1;
+        if (index < user.listLength) {
+            user.list[index].markComplete();
+            user.tasksIncomplete --;
+            System.out.println("\tAs you wish sir. I have marked this task as done:\n\t[X] " + user.list[index].item + "\n\tNow you have " + user.tasksIncomplete + " incomplete tasks");
```
teoziyiivy: Line length should be no longer than 120 chars.

src/main/java/Duke.java
```diff
@@ -1,10 +1,133 @@
...
+                "                           .,.   ,.                         \n" +
+                "                                                            \n" +
+                "                                                            \n" +
+                "      ,/,     ,//      //////*    //      */. */.    */////*\n" +
+                "      /@*    (@#@&            @@  .@&    %@*  #@*  &@       \n" +
+                "      /@*   #@* .@@    @@@@@@@@*   .@@  @@,   #@*  .@@@@@@@.\n" +
+                "      @@.  %@,    @@   @@   #@(      @@&@.    #@*         @&\n" +
+                " */**.    */.      /*  **    ./*      */      ,*.  ,******  ";
+        greet(logo);
+        Tasks user = new Tasks();
```
teoziyiivy: A different variable name can be used to show tasks is a plural collection of the user's tasks.

src/main/java/Duke.java
```diff
@@ -1,10 +1,133 @@
+import java.util.Random;
+import java.util.Scanner;
+
 public class Duke {
+    public static String addTaskSalutation(String item) {
```
teoziyiivy: printTaskSalutation might be a better method name.

src/main/java/Duke.java
```diff
@@ -1,10 +1,133 @@
...
+        default:
+            break;
+        }
+        return out;
+    }
+    public static void line() {
+        System.out.println("\t____________________________________________________________\n");
+    }
+
+    public static void showList(Tasks user) {
```
teoziyiivy: printList might be a better method name.

teoziyiivy (commented):
Good code overall! It's cool that your program has random responses, very interesting :)

src/main/java/Duke.java
```diff
@@ -1,10 +1,133 @@
...
+            out = "Sir, I can confirm that " + item + " has been added to your list of tasks\n";
+            break;
+        default:
+            break;
+        }
+        return out;
+    }
+    public static void line() {
+        System.out.println("\t____________________________________________________________\n");
+    }
```
kahhe: Consider naming as printLine() rather than line(). Methods' names should be verbs for easy understanding of their purposes.

src/main/java/Task.java
```diff
@@ -0,0 +1,18 @@
+public class Task {
+    protected String item;
+    protected boolean complete;
```
kahhe, line 6: Boolean variables should sound like a claim that can be true or false. isComplete may be more appropriate here.

kahhe (commented):
Code quality is quite well-adhered to. No deep nesting or methods with more than 30 LOC!

src/main/java/Deadline.java
```diff
@@ -0,0 +1,29 @@
+public class Deadline extends Task{
+    protected String by;
+    protected String originalInput;
+    public Deadline(String description, String by, String originalInput) {
```
stephlewyh: To improve code readability, you may wish to leave a line spacing the code block of class attributes and the constructor (`Deadline()`) here.

src/main/java/Duke.java
```diff
@@ -1,10 +1,240 @@
...
+import java.nio.file.Paths;
+
 public class Duke {
+    private static File save;
+
+    public static String addTaskSalutation(String item) {
+        Random rand = new Random();
+        String out = "";
+        switch(rand.nextInt(3)) {
+        case 0:
```
stephlewyh: You are using integers for the switch cases. How about creating more meaningful names for each case?

src/main/java/Duke.java
stephlewyh: It would help to add comments for your method to clarify what it is doing.

src/main/java/Duke.java
```diff
@@ -1,10 +1,240 @@
...
+    }
+
+    public static void showList(Tasks user) {
+        for (int i = 0; i < user.listLength; i++) {
+            System.out.println("\t" + (i + 1) + ". " + user.list[i] + "\n");
+        }
+    }
+    public static void addTodo(String input, Tasks user) {
+        user.list[user.listLength] = new Todo(input.substring(5));
+        user.listLength ++;
```
stephlewyh: The increment operator `++` is conventionally used without character spacing e.g.`listLength++`. You may wish to edit similar lines of code like this.

src/main/java/Duke.java
```diff
@@ -1,10 +1,240 @@
...
+    public static void line() {
+        System.out.println("\t____________________________________________________________\n");
+    }
+
+    public static void showList(Tasks user) {
+        for (int i = 0; i < user.listLength; i++) {
+            System.out.println("\t" + (i + 1) + ". " + user.list[i] + "\n");
+        }
+    }
+    public static void addTodo(String input, Tasks user) {
```
stephlewyh: You may wish to consider renaming the `user` parameter. It is strange that the parameter is `user` of type `Tasks`, when a task usually refers to object/item rather than a person.

src/main/java/Duke.java
```diff
@@ -1,10 +1,240 @@
...
+        user.list[user.listLength] = new Todo(input.substring(5));
+        user.listLength ++;
+        user.tasksIncomplete ++;
+    }
+    public static void addDeadline(String input, Tasks user) {
+        user.list[user.listLength] = new Deadline(input.substring(9, input.indexOf("/") - 1), input.substring(input.indexOf("/") + 1), input.substring(9));
+        user.listLength ++;
+        user.tasksIncomplete ++;
```
stephlewyh: You have repeated uses of `user.list[x]` and other variables that keep track of the array position to fill. This makes the code slightly more complicated than necessary.

You could simplify the code further by doing without the array index (in this case `listLength`).  
Try `ArrayList` so that you could add items to a growing list.

src/main/java/Duke.java
```diff
@@ -1,10 +1,240 @@
...
+            for (int i = index; i < user.listLength; i++) {
+                user.list[i] = user.list[i + 1];
+            }
+            user.listLength --;
+        } else {
+            throw new ArrayIndexOutOfBoundsException();
+        }
+    }
+
+    public static void response(String input, Tasks user) throws IllegalCommandException, IllegalTaskException, DueDateFormatException, MultiMarkDoneException, IOException, DeleteTaskFormatException {
```
stephlewyh: This method is rather long [>30 LoC](https://nus-cs2113-ay2122s1.github.io/website/se-book-adapted/chapters/codeQuality.html#avoid-long-methods).  How about abstracting out the procedures of each case into methods? Also see the [SLAP principle](https://nus-cs2113-ay2122s1.github.io/website/se-book-adapted/chapters/codeQuality.html#slap-hard).

src/main/java/Duke.java
```diff
@@ -1,10 +1,240 @@
...
+        String[] inputArr = input.split(" ");
+        switch (inputArr[0]) {
+        case ("list"):
```
stephlewyh: The brackets `()` for each case are unnecessary - see [switch statement style](https://se-education.org/guides/conventions/java/basic.html#layout).

src/main/java/Duke.java
```diff
@@ -1,10 +1,240 @@
...
+                i++;
+            }
+        } catch (IOException | MultiMarkDoneException e) {
+            System.out.println("Something went wrong: " + e.getMessage());
+        }
+    }
+    private static void writeToFile(Tasks user) throws IOException {
+        FileWriter fw = new FileWriter(save);
+        for (int i = 0; i < user.listLength; i++) {
+            fw.write(user.list[i].getType() + " | " + user.list[i].getStatus() + " | " + user.list[i].getOriginalInput() +  System.lineSeparator());
```
stephlewyh: This line is rather long, and could trip up the reader. The guideline is to have [<120 chars](https://se-education.org/guides/conventions/java/basic.html#layout) per line. 

Refer to the[ java string format() method ](https://www.javatpoint.com/java-string-format)example for another way to format long strings.

src/main/java/Duke.java
```diff
@@ -1,10 +1,240 @@
+import java.util.Random;
+import java.util.Scanner;
+import java.io.File;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.io.FileNotFoundException;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+
 public class Duke {
```
stephlewyh, line 19: Generally, you could consider adding more comments to describe what your methods or classes are doing.

PR #189, opened 2021-09-01
[Kairos Koh] iP

kairoskoh:


src/main/java/Duke.java
```diff
@@ -1,10 +1,181 @@
...
+        PrintGoodbyeMessage();
+    }
+
+    private static void ActiveChat() {
+        boolean isBye = false;
+        boolean isEmptyList = true;
+        String input;
+        Scanner in = new Scanner(System.in);
+        int counter = 0;
+        Task[] list = new Task[100];
```
lamkwjonathan: Perhaps 'tasks' is a more appropriate name, since it is plural.

src/main/java/Duke.java
```diff
@@ -1,10 +1,181 @@
...
+        PrintWelcomeMessage();
+
+        // Active Chat
+        ActiveChat();
+
+        // Goodbye Message
+        PrintGoodbyeMessage();
+    }
+
+    private static void ActiveChat() {
```
lamkwjonathan: I think 'ActiveChat' should be 'activeChat' instead, since methods should be in camelCase.

src/main/java/Duke.java
```diff
@@ -1,10 +1,181 @@
...
+                ProcessList(counter, list);
+            } else if (input.contains("done") ) { //check if done
+                ProcessDone(input, list);
+            } else {
+                counter = ProcessTasks(input, counter, list); //process tasks
+            }
+        }
+    }
+
+    private static void CheckIsEmptyList(Task[] list) {
```
lamkwjonathan: 'checkIsEmptyList' for 'CheckIsEmptyList', or perhaps even just 'isEmptyList' for boolean naming.
Similar camelCase method naming issues in other places as well.

src/main/java/Duke.java
```diff
@@ -1,10 +1,181 @@
...
+        System.out.println("Unable to tick off list.");
+        System.out.println("List is empty. Time to get productive!");
+        System.out.println("--------------------");
+    }
+
+    private static void PrintAddedTaskMessage(Task task, int i) {
+        System.out.println("--------------------");
+        System.out.println("Got it. I've added this task: ");
+        System.out.println("[" + task.getLetter() + "] "
+                + "[" + task.getStatusIcon() + "] "
```
lamkwjonathan: nice use of wrapped lines!

lamkwjonathan (commented):
Overall quite nicely done. Just watch out for camelCase when naming methods.

src/main/java/Task.java
```diff
@@ -0,0 +1,37 @@
...
+
+    public Task(){
+        this.description = "";
+        isDone = false;
+    }
+    public Task(String description) {
+        this.description = description;
+        this.isDone = false;
+    }
+
```
justinfidelis: Perhaps it will be cleaner to override the `.toString()` method to return the entire `[T][ ] task name` string, instead of needing separate public `getLetter()` and `getStatusIcon()` methods.

```
public String toString() {
    String checkbox = isCompleted ? "[X]" : "[ ]";
    return checkbox + " " + taskName;
}
```

It is also possible to further override the function in the subclasses. 
Example for Deadline subclass
```
@Override
public String toString() {
    String checkbox = isCompleted ? "[X]" : "[ ]";
    return "[D]" + super.toString() + " " + "(by: " + by + ")";
}
```

src/main/java/Duke.java
```diff
@@ -1,10 +1,181 @@
...
+
+    private static void ProcessDone(String input, Task[] list, int donePos) {
+        String itemNumDone = input.substring(donePos + 5, donePos + 6);
+        int itemNum = Integer.parseInt(itemNumDone);
+        list[itemNum - 1].setDone();
+        PrintDoneMessage(list, itemNum);
+    }
+
+    private static int ProcessTasks(String input, int counter, Task[] list) {
+        if (input.contains("todo")) {
```
justinfidelis: It might be cleaner to replace the strings "todo", "deadline", etc. with  named constants of type `static final String`

src/main/java/Duke.java
```diff
@@ -1,10 +1,181 @@
...
+            counter += 1;
+            PrintAddedTaskMessage(newTask, counter);
+        } else {
+            System.out.println("Please specify tasks: todo, deadline or event");
+            System.out.println("Example - type in the following: todo read book");
+        }
+        return counter;
+    }
+
+    private static void PrintDoneMessage(Task[] list, int itemNum) {
```
justinfidelis: Nice use of specific PrintMessage methods to keep the overall code cleaner and easier to understand

src/main/java/Duke.java
```diff
@@ -1,10 +1,181 @@
...
+        System.out.println("--------------------");
+        System.out.println("Nice! I've marked this task as done:");
+        System.out.println( itemNum + ".[" + list[itemNum - 1].getStatusIcon() + "] " + list[itemNum - 1].getDescription() );
```
justinfidelis: Linked to the comment on overriding the `.toString()` method in `Task`. By doing so, this part of code can be simplified to 
```
System.out.println( itemNum + "." + list[itemNum - 1])
```

src/main/java/Duke.java
```diff
@@ -1,10 +1,181 @@
...
+            counter += 1;
+            PrintAddedTaskMessage(newTask, counter);
+        } else if (input.contains("deadline")) {
+            int donePos = input.indexOf("/"); //finds pos of '/'
+            String description = input.substring(9,donePos);
+            String date = input.substring(donePos + 4);
+            Deadline newTask = new Deadline(description,date);
+            list[counter] = newTask;
+            counter += 1;
+            PrintAddedTaskMessage(newTask, counter);
```
justinfidelis: It might be cleaner to refactor this into a separate method. 
One way is to have additional methods in the Duke class that handles the different Task subclasses. 
Another is to create a separate class `TaskManager` that stores the array of Tasks and has functions to add the different Task subclasses to keep the main Duke class clean. __(I think this will be the best approach)__
Another is to overload the subclass constructors to take in `String[]` or a similar data structure.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,186 @@
...
+
+        // Welcome Message
+        printWelcomeMessage();
+
+        // Active Chat
+        activeChat();
+
+        // Goodbye Message
+        printGoodbyeMessage();
+    }
```
ongweisheng: Note if the comments are necessary, else remove them if they are not

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,186 @@
...
+                processDone(input, list);
+            } else {
+                try {
+                    counter = processTasks(input, counter, list); //process tasks
+                } catch ( IllegalTaskException e ) {
+                    System.out.println("Please include /by for deadline and /at for event");
+                }
+            }
+        }
+    }
```
ongweisheng: Consider implementing a Parser class to parse user inputs instead of putting everything in main/Duke

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,186 @@
...
+            Event newTask = new Event(description,date);
+            list[counter] = newTask;
+            counter += 1;
+            printAddedTaskMessage(newTask, counter);
+        } else {
+            System.out.println("Please specify tasks: todo, deadline or event");
+            System.out.println("Example - type in the following: todo read book");
+        }
+        return counter;
+    }
```
ongweisheng: Consider creating TaskManger class to handle task manipulation as well as storing their data for your task list

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,186 @@
...
+                + "|____/ \\__,_|_|\\_\\___|\n";
+        System.out.println("Hello from\n" + logo);
+
+        System.out.println("--------------------");
+        System.out.println("Hello! I'm Duke");
+        System.out.println("What can I do for you?");
+        System.out.println(" ");
+        System.out.println("--------------------");
+    }
+}
```
ongweisheng: Consider creating a UI class to handle your print statements instead

src/main/java/duke/Event.java
```diff
@@ -0,0 +1,26 @@
...
+    public char getLetter() {
+        return LETTER;
+    }
+
+    public String getDate() {
+        return "(at: " + date + ")";
+    }
+
+
+}
```
ongweisheng: Consider creating packages for your Event, Todo classes etc.

ongweisheng (commented):
Generally well done, do take note if comments are unnecessary do remove them and see if you can abstract more of your code

PR #191, opened 2021-09-01
[Edwin] iP

yzhedwin:


src/main/java/Duke.java
```diff
@@ -1,10 +1,53 @@
...
+                int j = 1;
+                System.out.println(border);
+                for (Task item : items) {
+                    if (item != null) {
+                        System.out.println(j + ".[" + item.getStatusIcon() + "] " + item.getDescription());
+                        j++;
+                    }
+                }
+                System.out.println(border);
+
```
astralum: May not need an extra newline here

astralum (commented):
Good adherence to coding standards!

src/main/java/Duke.java
```diff
@@ -1,10 +1,99 @@
...
+    private static int taskCount = 0;
+    private static String logo = " _____  ___ _____\n"
+            + "|___  | | ||_____| \n"
+            + "   / /  | |   / / \n"
+            + "  / /   | |  / /  \n"
+            + " / /___ | | /_/__  \n"
+            + "|_____| | ||_____| \n";
+    private static String border = "____________________________________________________________\n";
+
+    public static void addedTaskMessage(Task task) {
```
huien77: does it need to be past tense?

src/main/java/Duke.java
```diff
@@ -1,10 +1,99 @@
...
-        String logo = " ____        _        \n"
-                + "|  _ \\ _   _| | _____ \n"
-                + "| | | | | | | |/ / _ \\\n"
-                + "| |_| | |_| |   <  __/\n"
-                + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
+        String line;
+        printStartMessage();
+
+        do {
```
huien77: maybe can move the while loop to a new method, and call that method instead, following single level of abstraction

src/main/java/Duke.java
```diff
@@ -1,10 +1,99 @@
...
+                System.out.println(border);
+                System.out.println("Here is your list");
+
+                for (Task item : items) {
+                    if (item != null) {
+                        System.out.print(j + ".");
+                        System.out.println(item);
+                        j++;
+                    }
+                }
```
huien77: can consider moving this to a new method, might look neater, and no need so many level of indentation

src/main/java/Duke.java
```diff
@@ -1,10 +1,99 @@
...
+                System.out.println(border);
+
+            } else if (line.length() > 4 && line.substring(0,4).contains("done")) {
```
huien77: avoid magic number like `4`, instead store them as constants

huien77 (commented):
I can see some deep nesting and I think it would be good to create more methods to solve that
I also find some magic numbers, it would be good to avoid them
Overall, very good

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,210 @@
...
+import java.io.FileWriter;
+import java.io.IOException;
+
+
+public class Duke {
+    private static final int FOUR = 4;
+    private static final int FIVE = 5;
+    private static final int SIX = 6;;
+    private static final int EIGHT = 8;
+    private static final int NINE = 9;
```
stephlewyh: These string constants to refer to an integer may seem redundant and self-explanatory. 
If you are working to replace them with magic numbers I would suggest more meaningful names.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,210 @@
...
+
+
+    public static void printStartMessage() {
+        System.out.println(logo);
+        System.out.println(border + "Hi bro, my name is Echo");
+        System.out.println("What do you want?\n" + border);
+        System.out.println("Type bye to exit\n" + border);
+    }
+    public static boolean isInvalid(CommandList task, String line , String key) {
+        if (!line.split(key)[1].trim().isEmpty()) {
```
stephlewyh: This expression is a little complicated - calling 3 methods and a negation. See how to improve the quality of expressions [here](https://nus-cs2113-ay2122s1.github.io/website/se-book-adapted/chapters/codeQuality.html#avoid-complicated-expressions).

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,210 @@
...
+            System.out.println("Save folder already exists.");
+        }
+        try {
+            if (saveFile.createNewFile()) {
+                System.out.println("Successfully created save file");
+            } else {
+                System.out.println("Save file already exists");
+            }
+        } catch (IOException e) {
+            e.printStackTrace();
```
stephlewyh: How about including the previous statements in the [ try catch block](https://se-education.org/guides/conventions/java/basic.html#layout)? After all, this method is doing file related operations and the program could `catch` file related exceptions.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,210 @@
...
+    }
+
+    private static void updateSave() {
+        Task[] saveLists = new Task[items.size()];
+        items.toArray(saveLists);
+        try {
+           FileWriter fw = new FileWriter(Duke.fileName);
+            for (Task saveList : saveLists) {
+                fw.write(saveList.toString().charAt(1) + "/");
+                fw.write(saveList.getStatus() + "/" + saveList.getDescription());
```
stephlewyh: For the parameters to the file writer, you could have declared variables for each for them to improve readability. E.g. `fileStatus = saveList.getStatus()`.

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,210 @@
...
+                case "T":
+                    task.addTodo(Duke.items, description);
+                    break;
+            }
+        }
+        task.loadTaskCount(items);
+        System.out.println("Save file successfully loaded");
+    }
+
+    public static String readCommand(Scanner in, CommandList task, DukeException error) {
```
stephlewyh: This method is parsing the command string, but it is not obvious how it is done with seemingly arbitrary constants like` FOUR FIVE EIGHT` etc. You could consider a more direct way of comparing strings and checking parameters.

src/main/java/duke/command/CommandList.java
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    public void printErrorMessage(DukeException error) {
+        error.printError(error.getErrorType());
+    }
+
+    public void executeCommand(ArrayList<Task> items, DukeException error, String input) {
+        String[] inputs;
+        switch (command) {
+            case CMD_TODO:
```
stephlewyh: This deviates from the switch case statement [layout conventions ](https://se-education.org/guides/conventions/java/basic.html#layout)- indentation for `case` is not needed.

src/main/java/duke/command/CommandList.java
```diff
@@ -0,0 +1,152 @@
...
+                if (endPosition > SEVEN) {
+                    String num = input.substring(dividerPosition, endPosition);
+                    int taskNum = Integer.parseInt(num) - 1;
+                    removeItem(items, taskNum);
+                }
+                break;
+            default:
+                printErrorMessage(error);
+                break;
+        }
```
stephlewyh: While you have organized different procedure into separate cases, the executeCommand method is still rather long. How about trying [SLAP principle](https://nus-cs2113-ay2122s1.github.io/website/se-book-adapted/chapters/codeQuality.html#slap-hard) here?

src/main/java/duke/Duke.java
```diff
@@ -0,0 +1,210 @@
...
+                fw.write("\n");
+            }
+            System.out.println("Automatically saved to " + fileName);
+            fw.close();
+        } catch (IOException e) {
+            e.printStackTrace();
+        }
+    }
+
+    private static void readSave(CommandList task) throws FileNotFoundException {
```
stephlewyh: It would be good to provide method descriptions as [comments/javadocs](https://se-education.org/guides/conventions/java/basic.html#comments) to improve code readability.

PR #193, opened 2021-09-01
[Minyue] iP

astralum:


src/main/java/Duke.java
```diff
@@ -56,7 +54,7 @@ public static void main(String[] args) {
                   int inputNum = Integer.parseInt(inputNumStr);
                   if (inputNum > 0 && inputNum <= Task.getNumOfTasks()) {
                       int id = inputNum - 1;
-                      tasks[id].setIsDone();
+                      tasks[id].setDone();
```
LS-Young: nice :D

src/main/java/Duke.java
```diff
@@ -1,10 +1,56 @@
+import java.util.Scanner;
+
 public class Duke {
-    public static void main(String[] args) {
+
+    public static void sayHello() {
         String logo = " ____        _        \n"
```
yuejunfeng0909: since logo should be a constant, can change the variable name to uppercase and set it as a constant

src/main/java/TaskHandler.java
```diff
@@ -0,0 +1,149 @@
+public class TaskHandler {
+
+    protected Task[] tasks;
+
+    public TaskHandler() {
+        this.tasks = new Task[100];
```
yuejunfeng0909: you may want to use a constant MAX_TASKS instead of putting a random number here

src/main/java/TaskHandler.java
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    public String handleTasks(String line) {
+        String lc = line.toLowerCase();
```
yuejunfeng0909: try not to use acronyms in the codes

yuejunfeng0909 (commented):
You may want to consider add the methods to the individual classes instead of using a separate function to handle all the methods

src/main/java/duke/Event.java
```diff
@@ -0,0 +1,16 @@
+package duke;
+
+public class Event extends Task {
+
+    protected String at;
+
+    public Event(String description, String at) {
```
shivin9: Can have a more descriptive name than 'at'

src/main/java/duke/Formatter.java
```diff
@@ -0,0 +1,38 @@
...
+
+    public static String returnInputStart() {
+        return INPUT_START;
+    }
+
+    public static String returnOutputStart() {
+        return OUTPUT_START;
+    }
+    
+    public static void printInputStart() {
```
shivin9: You can have a small description before functions...

src/main/java/duke/TaskHandler.java
```diff
@@ -0,0 +1,275 @@
...
+                return returnEventNoAt();
+            } else {
+                return addEvent(line);
+            }
+        } else {
+            return returnInputInvalid();
+        }
+    }
+
+    private void openFile() {
```
shivin9: Comments before functions would help...

src/main/java/duke/TaskHandler.java
```diff
@@ -0,0 +1,275 @@
...
+
+    public String addTodo(String line) throws IllegalArgumentException {
+        if (line.length() <= 5) {
+            throw new IllegalArgumentException(EMPTY_DESCRIPTION_MSG);
+        }
+        String description = line.substring(5).trim();
+        Todo newTodo = new Todo(description);
+        tasks.add(newTodo);
+        appendLinetoFileData(newTodo.toString());
+        return returnAddTaskSuccess() + newTodo.toString();
```
shivin9: You can break this line into two according to me

src/main/java/duke/TaskHandler.java
```diff
@@ -0,0 +1,275 @@
...
+        } else if (inputIsTodo(lc)) {
+            return addTodo(line);
+        } else if (inputIsDeadline(lc)) {
+            if (!deadlineContainsBy(lc)) {
+                return returnDeadlineNoBy();
+            } else {
+                return addDeadline(line);
+            }
+        } else if (inputIsEvent(lc)) {
+            if (!eventContainsAt(lc)) {
```
shivin9: Try to avoid double nesting if possible...

src/main/java/duke/Task.java
```diff
@@ -0,0 +1,38 @@
+package duke;
+
+public class Task {
+
+    protected String description;
+    protected boolean isDone;
+
```
shivin9: this line can be removed IMHO

shivin9 (commented):
Nice code overal. Formatting can be imporved though

PR #197, opened 2021-09-02
[Xue Yushan] iP

Mist0919:


src/main/java/list.java
```diff
@@ -0,0 +1,42 @@
+public class List {
+    private static String[] currentList = new String[100];
+    public static int listSize = 0;
+    private static int[] isDone = new int[100];
```
itsleeqian: name of this array "int[] isDone" sounds like a boolean name. 
maybe change the name to better represent what you are trying to store

src/main/java/List.java
```diff
@@ -0,0 +1,42 @@
...
+                System.out.print("] " + currentList[i-1] + "\n");
+            }
+        }
+    }
+
+    public static void mark(int number) {
+        isDone[number-1] = 1;
+    }
+
+    public static void markedItems() {
```
itsleeqian, line 35: name of this function "markedItems()" sounds like a list. maybe should make it sound like a verb instead

src/main/java/Duke.java
```diff
@@ -6,5 +9,49 @@ public static void main(String[] args) {
...
+        Scanner input = new Scanner(System.in);
+        while (true) {
+            String str = input.nextLine();
+            if (str.equals("bye")) {
+                System.out.println("____________________________________________________________\n"
+                        + "Bye. Hope to see you again soon!\n"
+                        + "____________________________________________________________\n");
+                break;
+            }
+            else if (str.equals("list")) {
```
theodorekwok: The else if should be on line 25, to follow the K&R egyption style bracketing, likewise for line 32.
It was done correctly on line 52

src/main/java/Duke.java
```diff
@@ -1,3 +1,6 @@
+import java.util.Scanner;
+import java.lang.String;
+
 public class Duke {
     public static void main(String[] args) {
```
theodorekwok: The main method exceeds 30 lines, consider refactoring the console input and output into separate methods

src/main/java/List.java
```diff
@@ -0,0 +1,42 @@
+public class List {
+    private static String[] currentList = new String[100];
+    public static int listSize = 0;
+    private static int[] isDone = new int[100];
```
theodorekwok, line 4: The value 100 is a magic number, consider extracting it as a constant variable

src/main/java/Duke.java
```diff
@@ -6,5 +9,49 @@ public static void main(String[] args) {
...
+                    System.out.println("____________________________________________________________\n"
+                            + "The index is out of range.\n"
+                            + "____________________________________________________________\n");
+                } else {
+                    List.mark(index);
+                    System.out.println("____________________________________________________________\n"
+                            + "Nice! I've marked this task as done: ");
+                    List.markedItems();
+                    System.out.println("____________________________________________________________");
+                }
```
theodorekwok: The abstraction levels here are different violating the single level of abstraction principle (SLAP) from the for loop in line 34 to 39 and the method calling in line 46 and 49 respectively.
Consider abstracting the for loop in line 34 to 39 as a separate method that is responsible for the console input.

src/main/java/list.java
```diff
@@ -0,0 +1,42 @@
...
+        currentList[listSize] =  input;
+        listSize++;
+        System.out.println("____________________________________________________________\n"
+                + "Added: " + input + "\n"
+                + "____________________________________________________________\n");
+    }
+
+    public static void printItems() {
+        if (listSize == 0) {
+            System.out.println("The list is empty.");
```
theodorekwok, line 41: The if and else makes the happy path not prominent. 
```suggestion
            System.out.println("The list is empty.");
            return;
```

src/main/java/List.java
```diff
@@ -0,0 +1,42 @@
...
+                    System.out.print("X");
+                } else {
+                    System.out.print(" ");
+                }
+                System.out.print("] " + currentList[i-1] + "\n");
+            }
+        }
+    }
+
+    public static void mark(int number) {
```
theodorekwok, line 31: Consider renaming this method to make it more clear on what it does. For eg setTaskAsCompleted

theodorekwok (commented):
Overall good job following most of the code guidelines. Do consider refactoring the functionalities into methods to reduce the number of conditionals

src/main/java/Duke.java
```diff
@@ -12,52 +10,48 @@ public static void main(String[] args) {
...
-            else if (str.equals("list")){
-                System.out.println("____________________________________________________________\n" +
-                        "Here are the tasks in your list: ");
-                list.printItems();
+            else if (str.equals("list")) {
+                System.out.println("____________________________________________________________\n"
+                        + "Here are the tasks in your list: ");
+                List.printItems();
                 System.out.println("____________________________________________________________");
             }
```
nguyin: The bracket for the if-else statement should be on the same line as else, so the line should be : 
} else {

src/main/java/Duke.java
```diff
@@ -12,52 +10,48 @@ public static void main(String[] args) {
...
-                "Hello! I'm Duke\n" +
-                "What can I do for you?\n" +
-                "____________________________________________________________\n");
+        System.out.println("____________________________________________________________\n"
+                + "Hello! I'm Duke\n"
+                + "What can I do for you?\n"
+                + "____________________________________________________________\n");
         Scanner input = new Scanner(System.in);
-        while (true){
+        while (true) {
```
nguyin: I like that you added the spacing after for the while loop and the { bracket

src/main/java/Duke.java
```diff
@@ -1,10 +1,289 @@
...
+    public static final int EVENT_POS = 6;
+    public static final int TYPE_POS = 4;
+    public static final int TASK_POS = 10;
+
     public static void main(String[] args) {
+
+        greeting();
+        ArrayList<Task> tasks = getDatafromFile("tasks");
+
+        if (tasks.size() == 0) {
```
iamabhishek98: You can also do something like:
```
if (!tasks.size()) {
}
```

src/main/java/Duke.java
```diff
@@ -1,10 +1,289 @@
...
+                    }
+                    if (type == "") {
+                        System.out.println(line
+                                + "☹ OOPS!!! I'm sorry, but I don't know what that means :-(\n"
+                                + line);
+                    } else {
+                        System.out.println(line
+                                + "☹ OOPS!!! The description of a " + type + " cannot be empty.\n"
+                                + line);
+                    }
```
iamabhishek98: Avoid using so many nested try catch blocks. Please abstract out the logic into different functions

src/main/java/Duke.java
```diff
@@ -1,10 +1,289 @@
...
+                exit();
+                break;
+            } else if (inputString.equals("list")) {
+                try {
+                    printList(tasks);
+                } catch (DukeException d) {
+                    System.out.println(line
+                            + d.getMessage() + "\n" + line);
+                }
+            } else if (inputString.contains("done") == true) {
```
iamabhishek98: There is no need to explicitly check if it is equal to true because the contains method returns a boolean value. You can change this to:
```
else if (inputString.contains("done")) {
}
```

src/main/java/Duke.java
```diff
@@ -1,10 +1,289 @@
...
+            if (reader != null) {
+                try {
+                    reader.close();
+                } catch (IOException e) {
+                    e.printStackTrace();
+                }
+            }
+        }
+        return tasks;
+    }
```
iamabhishek98: Having such long functions can be hard to read. Please consider abstracting out repeated code in line with the DRY (Don't Repeat Yourself) principle. Also add comments to improve the readability of your code

src/main/java/List.java
iamabhishek98: Agreed! Also make sure the constants are named using proper naming conventions.

src/main/java/List.java
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    public static void printItems() {
+        if (listSize == 0) {
+            System.out.println("The list is empty.");
+        } else {
+            for (int i = 1; i <= listSize; i++) {
+                System.out.print(i + ". " );
+                System.out.print("[");
+                if (isDone[i-1] == 1) {
```
iamabhishek98, line 21: Please maintain spacing between the operators and the operands

src/main/java/Task.java
```diff
@@ -0,0 +1,19 @@
...
+    protected String description;
+    protected String isDone;
+
+    public Task(String description) {
+        this.description = description;
+        this.isDone = " ";
+    }
+
+    public final void markedAsDone() {
+        this.isDone = "X";
```
iamabhishek98, line 11: As mentioned earlier, avoid using magic strings - please extract them out as constants.

src/main/java/list.java
```diff
@@ -0,0 +1,65 @@
...
+        }
+    }
+
+    public static void addToList(String input) {
+        if (input.contains("todo")) {
+            l[listSize] = input.substring(5);
+            type[listSize] = "T";
+        } else if (input.contains("deadline")) {
+            String by = input.substring(input.indexOf("/") + 4);
+            l[listSize] = input.substring(9, input.indexOf("/")) + " (by: " + by
```
iamabhishek98, line 20: Avoid having complex nested logic like this. Either separate it out into different lines or abstract it out into different functions

iamabhishek98 (commented):

PR #210, opened 2021-09-14
Week 6 jar

Poopies99:


src/main/java/Duke.java
```diff
@@ -1,10 +1,77 @@
...
+                }
+            } else if (line.startsWith("event")) {
+                try {
+                    processManager.handleEventRequest(line);
+                } catch (EventException e) {
+                    e.printStatement();
+                    System.out.println("Invalid Event Request. Format: event (description) /at (Date)");
+                    System.out.println("    ____________________________________________________________\n");
+                }
+            } else {
```
EthanWong2212, line 71: Perhaps it would be better to use constant variables to avoid use of magic strings.
Would also be good to shorten the method as it is very long which thus affects quality

src/main/java/ProcessManager.java
```diff
@@ -0,0 +1,290 @@
...
+import tasks.Task;
+import tasks.Todo;
+import java.util.ArrayList;
+import java.io.File;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.Scanner;
+
+public class ProcessManager {
+    /* ---- CONSTANTS ---- */
```
EthanWong2212, line 18: Since the code is self explanatory perhaps it would be better to refrain from repeating description in comment

src/main/java/ProcessManager.java
```diff
@@ -0,0 +1,290 @@
...
+    }
+    /* ---- ------------- ---- */
+
+    /* ---- Handle Functions ---- */
+    public void handleEventRequest(String line) throws EventException {
+        if (!line.contains(EVENT_KEYWORD)) {
+            throw new EventException("Event Request Does Not Contain /at");
+        }
+        Integer dividerPosition = line.indexOf(EVENT_KEYWORD);
+        if (dividerPosition - TASK_DESCRIPTION_DIVIDER - EVENT_DIVIDER < 0) {
```
EthanWong2212, line 122: Perhaps it would be better to put such operations in a variable to improve readability

src/main/java/ProcessManager.java
```diff
@@ -0,0 +1,290 @@
...
+        }
+    }
+    /*--- -------- --- */
+
+    /*--- Messages --- */
+    public String taskAddedMessage() {
+        return LINE_DIVIDER + ADD_TASK_REPLY
+                + GAP + taskList.get(getLastIndex()).toString() + "\n"
+                + getAddTaskReturn(getLastIndex() + ARRAY_INDEX_FINDER) + LINE;
+    }
```
EthanWong2212, line 245: Maybe it would be better to make such methods print the statements instead of adding a println in the function where method is called

src/main/java/tasks/Task.java
```diff
@@ -0,0 +1,37 @@
...
+    protected boolean isDone;
+
+    public Task(String description) {
+        this.description = description;
+        this.isDone = false;
+    }
+
+    public String toString() {
+        return getStatusIcon() + getDescription();
+    }
```
EthanWong2212, line 16: Maybe this should be override

EthanWong2212 (commented):
Overall good job, keep up the good work

src/main/java/Duke.java
Poopies99: Hi Ethan, when you say shorten the method, do you mean shorten the number of characters of the method name?
So for example change 'processManager.handleDeadlineRequest' to something like 'processManager.addDeadline'?

src/main/java/tasks/Task.java
Poopies99: Hi Ethan, Task is an abstract class so wouldn't the method of toString() within task be 'Override' by other classes extending on it?

src/main/java/Duke.java
EthanWong2212: I mean you should avoid having too many lines of code in a single method
https://nus-cs2113-ay2122s1.github.io/website/se-book-adapted/chapters/codeQuality.html#avoid-long-methods

src/main/java/tasks/Task.java
EthanWong2212: toString() is an inbuilt java object method so you are overriding it when you invoke this method (which you did in deadline class).

Poopies99: Alright noted

